PR #221, opened 2026-05-19
Spec: external command sources and dispatchers

s0undt3ch:
Design document only — no implementation in this PR.

## Summary
Lets toolr expose commands the user never wrote (Django management commands, Jenkins jobs, etc.), with first-class completion and \`--help\`, by:
- A *source plugin* contract (\`scan(root, config) -> SourceFragment\`).
- An *attachment* config in \`tools/pyproject.toml\` (\`[tool.toolr.sources.<name>]\` + \`[[attach]]\`).
- A *dispatcher* convention: any user toolr command with a \`toolr.sources.DispatchCommand\`-annotated parameter receives matched-child info from the runtime.
- Reference plugin \`toolr-django\` lives in-tree at \`crates/toolr-django/\` with the existing CI matrix; only the release action is independent (\`release-toolr-django.yml\`).

Full design at \`specs/2026-05-19-external-command-sources-design.md\`.

## Open after merge
- Implementation plan (covered by the writing-plans hand-off).
- The deferred edge cases listed under \"Open questions\" / \"Migration notes\" in the spec.

PR #222, opened 2026-05-19
feat(argparse): built-in argparse scanner + DispatchCommand dispatcher contract

s0undt3ch:
Implements `specs/2026-05-19-external-command-sources-design.md` and Plan A at `specs/2026-05-19-external-command-sources-plan-a.md`.

## What's in the box

- **`toolr.sources` Python module** — public types: `ArgSchema`, `CommandSchema`, `DispatchCommand` (+ `.argv` reconstruction).
- **Dispatcher detection** — annotation-driven via `dispatched: DispatchCommand` keyword-only parameter; lives in `toolr.utils._signature.detect_dispatch_parameter`.
- **Manifest field** — `dispatched_from: Option<String>` on Rust `Command` (no schema version bump; pre-1.0).
- **Argparse AST scanner** — `crates/toolr-core/src/argparse/`: TOML config, ruff_python_parser-based AST walk with signature-shape filter (any receiver name, but call shape must match argparse), glob-expanded scan paths, common_args hoisting, attachment validation, child grafting, end-to-end `run_for_project` orchestrator.
- **Static-manifest integration** — `build_static_manifest` runs the scanner and grafts children alongside user-decorated commands.
- **Parser support for dispatchers** — `SourcesImports` tracks `from toolr.sources import DispatchCommand` (and aliases / qualified forms) so the static parser skips the dispatcher's reserved kwarg when extracting CLI args.
- **Runtime dispatch** — dispatch.rs branches on `dispatched_from`, packs the matched leaf's args, builds an `ExecutionSpec` with a `dispatch` payload (the toolr binary doesn't link pyo3; the JSON spec is the existing wire format). The Python runner constructs the `DispatchCommand` and calls `invoke_dispatcher(ctx, func, parent_kwargs, child_name, child_args, child_schema)`, which calls `func(ctx, **parent_kwargs, dispatched=DispatchCommand(...))`.

## Tests

- 21 sources/argparse-related test files cover the schema types, detection, scanner, validation, grafting, runtime injection.
- 4 E2E tests under `tests/sources/test_e2e.py`:
  - explicit `toolr project manifest rebuild` + dispatch via sidecar JSON ✅
  - same source attached to two parents ✅
  - collision across sources fails build ✅
  - auto-rebuild on missing manifest — **skipped**: not implemented in `main.rs` today; `load_or_empty` silently returns an empty manifest so clap rejects the user's command before any rebuild fires. Tracked as follow-up.
- All cargo + pytest + clippy green.

## Deferred / follow-up

- **Auto-rebuild on missing manifest**: needs main.rs to detect missing manifest and run `rebuild_manifest_full` before clap parsing.
- **Dispatcher's own CLI flags reachable via grafted children**: the spec's `toolr jenkins job --cpu 5000m migrate` invocation pattern requires the dispatcher command to host children as subcommands. Today grafted children are siblings of the dispatcher in the parent group, so the user invokes `toolr jenkins migrate ...` and the dispatcher's --cpu/--ram flags aren't reachable on that path. A separate task to merge dispatcher args into the parent group's clap command.
- **`toolr-django` reference plugin** is Plan B (still pending — argparse scanner now covers most of the Django case, so Plan B's scope is much narrower).

### DIFF
--- a/crates/toolr-core/src/argparse/attach.rs
+++ b/crates/toolr-core/src/argparse/attach.rs
@@ -1,0 +1,1 @@
+//! Attachment validation + grafting — implemented in Task 12.
--- a/crates/toolr-core/src/argparse/attach.rs
+++ b/crates/toolr-core/src/argparse/attach.rs
@@ -1,1 +1,169 @@
-//! Attachment validation + grafting — implemented in Task 12.
+//! Graft scanned commands under parent dispatchers, with validation.
+
+use std::collections::HashMap;
+
+use thiserror::Error;
+
+use crate::argparse::config::ArgparseBlock;
+use crate::manifest::Command;
+
+#[derive(Debug, Error)]
+pub enum AttachError {
+    #[error("source {block:?} attaches to unknown parent {parent:?}{hint}")]
+    UnknownParent {
+        block: String,
+        parent: String,
+        hint: String,
+    },
+    #[error("parent {parent:?} has no DispatchCommand-annotated keyword parameter")]
+    NotADispatcher { parent: String },
+    #[error(
+        "child name collision on parent {parent:?}: {name:?} is provided by both {a:?} and {b:?}"
+    )]
+    Collision {
+        parent: String,
+        name: String,
+        a: String,
+        b: String,
+    },
+}
+
+pub fn validate_attachments(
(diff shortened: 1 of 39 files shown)


s0undt3ch:

- #221
    - #222 ◀

PR #223, opened 2026-05-20
complete: drop spurious `--` in fish completion script

s0undt3ch:
## Summary

`set -a args -- $current` in the embedded fish completion script does not behave as an end-of-options marker. Fish's `set` builtin appends `--` as a literal value, so the generated call became:

```
toolr __complete "$PWD" <tokens> -- ""
```

clap consumed the `--` as its options-terminator and dropped the trailing empty token, so `toolr <group> <TAB>` (and `toolr <group> <leaf> <TAB>`) returned no candidates. Only the top-level `toolr <TAB>` worked because no extra tokens followed.

## Fix

Remove the `--`. Fish's `set` does not interpret the in-progress word as a flag, so the separator was never needed for safety either.

## Test plan

- [x] `cargo test -p toolr-core complete::` passes (39/39)
- [x] Manual repro: `fish -c 'complete --do-complete "toolr ci "'` now returns subcommands
- [x] Added regression test `fish_script_does_not_inject_literal_dash_dash_into_args`

### DIFF
--- a/crates/toolr-core/src/complete/scripts/fish.fish
+++ b/crates/toolr-core/src/complete/scripts/fish.fish
@@ -11,7 +11,11 @@ function __toolr_complete
     set -l current (commandline -ct)
     # Drop the leading `toolr` token.
     set -l args $tokens[2..-1]
-    set -a args -- $current
+    # Append the in-progress word as the trailing token. Do NOT use `--`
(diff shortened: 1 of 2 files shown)


s0undt3ch:

- #223 ◀

PR #224, opened 2026-05-20
complete: ship completion for built-in `self` and `project` commands

s0undt3ch:
## Summary

The completion engine in `toolr_core::complete` is purely manifest-driven, so it only saw commands harvested from a project's `tools/` directory. The binary's own `self` and `project` subcommand trees never appeared as tab candidates.

This PR adds a binary-side `builtin_completions` module that produces synthetic `Group` + `Command` entries mirroring `cli::build_command`'s built-in tree. `run_complete` merges them into the loaded manifest before delegating to `serve_completions`. The synthetic entries are never invoked at dispatch time — `dispatch` intercepts `self` / `project` before any manifest lookup happens.

## Coverage

- Top-level: `toolr <TAB>` now offers `self` and `project` alongside user commands
- `toolr self <TAB>` → `build-manifest`, `cache`, `completion`
- `toolr project <TAB>` → `init`, `deps`, `venv`, `manifest`
- All leaf subcommands (`self cache prune`, `project venv path`, etc.)
- Flags for `self cache prune` (`--all`, `--stale-after-days`, `--dry-run`, `--yes`)
- Enum values for `self completion {install,print} <shell>` (bash/zsh/fish)
- Enum values for `project init --venv-location` (cache/in-tree)

## Maintenance

When `cli::build_command` grows a new built-in subcommand, mirror the change in `builtin_completions.rs`. The module's doc comment calls this out; future work could derive the synthetic entries from the clap tree directly.

## Test plan

- [x] 7 new unit tests in `builtin_completions::tests` exercise serve_completions against the merged manifest
- [x] Manual: `toolr __complete "$PWD" self ""` → `build-manifest\ncache\ncompletion`
- [x] Manual: `toolr __complete "$PWD" self cache prune --` → `--all --dry-run --stale-after-days --yes`
- [x] `cargo clippy -p toolr` clean

## Stacks on

#223 (fish completion separator fix)

### DIFF
--- a/crates/toolr/src/builtin_completions.rs
+++ b/crates/toolr/src/builtin_completions.rs
@@ -1,0 +1,290 @@
+//! Synthetic manifest entries for toolr's own built-in subcommands.
+//!
+//! The completion engine in [`toolr_core::complete`] is purely
+//! manifest-driven — it knows nothing about the binary's own subcommands
+//! (`self`, `project`). This module supplies the entries that mirror
+//! [`crate::cli::build_command`] so the engine can offer them as
+//! candidates. The synthetic [`Command`] entries are never invoked; only
+//! their `name`, `group`, `arguments[].name`, `arguments[].kind`, and
+//! `arguments[].allowed_values` matter for completion classification.
+//!
+//! When [`crate::cli::build_command`] grows or changes a built-in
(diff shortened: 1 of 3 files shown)


s0undt3ch:

- #223
    - #224 ◀

PR #225, opened 2026-05-20
complete: honour `--force` literally + distinguish install outcomes

s0undt3ch:
## Summary

Previously `install_script` short-circuited to `AlreadyInstalled` whenever the on-disk content matched the embedded payload, **regardless of `--force`**. The check happened before `--force` was even consulted:

```rust
if let Ok(existing) = std::fs::read_to_string(&path) {
    if existing == payload {
        return Ok(InstallOutcome::AlreadyInstalled { path });   // shortcuts before --force check
    }
    if !opts.force {
        return Ok(InstallOutcome::SkippedNeedsForce { path });
    }
}
```

This bit users running a stale binary that shipped a buggy completion script: the bug got written to disk, then refreshing with the same (still buggy) binary said "already installed" without writing, even with `--force`. The bug source was hiding itself on both sides of the equality check.

## Behaviour change

`--force` now writes unconditionally. When the on-disk content already matched the new payload, the outcome distinguishes the no-op-but-forced rewrite from a fresh install via a new `PriorState` discriminant on `InstallOutcome::Wrote`. The dispatcher renders five distinct messages:

| --force | on disk | outcome | message |
|---|---|---|---|
| any | absent | `Wrote { prior: None }` | `wrote {shell} completion script to {path}` |
| true | identical | `Wrote { prior: Identical }` | `rewrote {shell} completion script at {path} (--force; content unchanged)` |
| true | differed | `Wrote { prior: Differed }` | `replaced existing {shell} completion script at {path}` |
| false | identical | `AlreadyInstalled` | `{shell} completion already up to date at {path}` (was: "already installed") |
| false | differed | `SkippedNeedsForce` | `refusing to overwrite {path} (use --force to replace)` |

## Test plan

- [x] New `install_with_force_overwrites_even_when_content_matches` regression test covers the stale-binary scenario.
- [x] Existing install tests updated to assert the new `PriorState` discriminant.
- [x] All 40 completion tests pass.
- [x] Manual: `toolr self completion install fish` → up-to-date message; with `--force` → rewrote-content-unchanged message.

## Stacks on

#224 (built-in command completions), which stacks on #223 (fish script fix)

### DIFF
--- a/crates/toolr-core/src/complete/install.rs
+++ b/crates/toolr-core/src/complete/install.rs
@@ -27,9 +27,28 @@ pub struct InstallOptions {
 
 #[derive(Debug)]
 pub enum InstallOutcome {
-    Wrote { path: PathBuf },
+    /// The embedded script is now on disk at `path`. `prior` describes
+    /// what (if anything) was there before the write so the caller can
+    /// distinguish a fresh install, a forced re-write of identical
+    /// content, and a forced replacement of differing content.
+    Wrote { path: PathBuf, prior: PriorState },
+    /// `--force` was not set and the on-disk content already matched
+    /// the embedded payload; nothing was written.
     AlreadyInstalled { path: PathBuf },
+    /// `--force` was not set and the on-disk content differs from the
+    /// embedded payload; nothing was written.
     SkippedNeedsForce { path: PathBuf },
+}
+
+#[derive(Debug)]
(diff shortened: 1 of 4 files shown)


s0undt3ch:

- #223
    - #224
        - #225 ◀

PR #226, opened 2026-05-20
cache: backfill meta.json on touch when sidecar is missing

s0undt3ch:
## Summary

`toolr project venv path` reports a venv path purely from `resolve_venv_path` (filesystem-derived, no existence check). `toolr self cache list` walks the cache root but [`enumerate_caches`](crates/toolr-core/src/cache/enumerate.rs#L67-L69) silently skips entries without a `meta.json` sidecar. Cache entries created by older toolr binaries — or any code path that produced a venv without writing the sidecar — therefore stay invisible to `list` / `prune` forever.

Reproduction:

```
$ toolr project venv path
/Users/pedro.algarvio/Library/Caches/toolr/f17b14b02ad081cd/venv
$ toolr self cache list
toolr: no cached virtualenvs found under /Users/pedro.algarvio/Library/Caches/toolr
```

Existing on disk:

```
~/Library/Caches/toolr/f17b14b02ad081cd/
└── venv/             # no meta.json
```

`touch_last_used` had the right shape to self-heal — it's called on every user-command dispatch — but it deliberately treated a missing sidecar as a noop because it had no context to write one from.

## Fix

Add `touch_or_backfill(cache_dir, repo_path, toolr_version, python_version)` next to `touch_last_used`:

- existing sidecar → update `last_used_at` (old behaviour)
- missing sidecar → synthesise a fresh `meta.json` from the supplied context (new)

Dispatch already has all three values at the call site — `resolve_venv_path` returns `python_version` — so plumbing is just keeping that field through the unpacking. The previous bind `(python, venv_dir)` becomes `(python, venv_dir, python_version)`.

The next user-command invocation against an unmetadata'd cache entry writes the sidecar; subsequent `cache list` calls then see it without any manual intervention. `touch_last_used` is preserved as the low-level update-only entry point.

## Caveats

- `created_at` on backfilled entries reflects the time of backfill, not the time the venv directory was actually materialised — the real birth time is not recoverable from on-disk state. Documented inline.
- Backfill only runs on a real *user-command* dispatch (the existing touch path). Running only `self` or `project` subcommands doesn't trigger it, but those are the very commands where the missing-sidecar UX bites; running any tools command once fixes both surfaces.

## Test plan

- [x] New `touch_or_backfill_writes_sidecar_when_missing` regression test
- [x] New `touch_or_backfill_updates_existing_sidecar` (existing-sidecar context is ignored)
- [x] Existing `touch_last_used_is_a_noop_when_sidecar_is_missing` updated to assert it does NOT create a sidecar (negative invariant)
- [x] Manual: `toolr version current` on my repo created the missing `meta.json`; `toolr self cache list` then showed the entry

## Stacks on

#225 (--force install fix), which stacks on #224 (built-in command completions), which stacks on #223 (fish script fix)

### DIFF
--- a/crates/toolr-core/src/cache/mod.rs
+++ b/crates/toolr-core/src/cache/mod.rs
@@ -13,7 +13,7 @@ pub use enumerate::{CachedVenv, dir_size_bytes, enumerate_caches};
 pub use hint::{HintConfig, compute_hint};
 pub use init::write_meta_for_new_venv;
 pub use meta::{Meta, MetaError, SCHEMA_VERSION};
-pub use touch::touch_last_used;
+pub use touch::{touch_last_used, touch_or_backfill};
 
 #[cfg(test)]
 mod tests;
--- a/crates/toolr-core/src/cache/tests.rs
+++ b/crates/toolr-core/src/cache/tests.rs
@@ -103,7 +103,7 @@ fn write_meta_for_new_venv_overwrites_existing() {
     assert_eq!(loaded.python_version, "3.13.0");
 }
 
-use super::touch::touch_last_used;
+use super::touch::{touch_last_used, touch_or_backfill};
 
 #[test]
 fn touch_last_used_updates_only_last_used_at() {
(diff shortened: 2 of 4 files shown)


s0undt3ch:

- #223
    - #224
        - #225
            - #226 ◀

PR #230, opened 2026-05-20
bootstrap: rebuild manifest on `--help` and bare `toolr` + re-export DispatchCommand

s0undt3ch:
## Summary

Two small fixes:

### `bootstrap`: rebuild manifest on `--help` and bare `toolr`

`toolr --help` rendered only the built-in tree (`project`, `self`) and silently hid every user group whenever `tools/.toolr-manifest.json` was missing. The user had to run a real command first to discover the rebuild path existed — a UX trap, especially right after `project init` or after wiping caches.

Drop `--help` / `-h` and the bare-`toolr` branch from `should_skip_auto_rebuild` in `crates/toolr/src/bootstrap.rs`. Keep the other skips:

- `--version` / `-V` — binary metadata only, manifest-independent
- `__complete` — Tab completion must stay silent and fast
- `self` / `project` / `init` — built-in subcommands that don't need the user's manifest to run

The pyproject.toml-existence gate (already in place) still scopes auto-rebuild to projects that actually opted into toolr — `--help` outside a toolr project is unaffected.

### `toolr-py`: re-export `DispatchCommand` from package root

`DispatchCommand` is the public contract for argparse-style dispatcher targets, but it was only reachable as `from toolr.sources import DispatchCommand` while its siblings (`Context`, `command`, `arg`, ...) all re-export from the package root. Add it to `__all__` and the top-level imports so dispatcher annotations match the rest of the public API surface:

    from toolr import DispatchCommand

## Test plan

- [x] `cargo test -p toolr --bin toolr bootstrap` — 11 tests pass, including 3 new `fires_for_*` cases
- [x] `--version` / `-V` still skip the rebuild (asserted)
- [x] `__complete`, `self`, `project`, `init` still skip the rebuild (asserted)
- [x] Manual: `toolr --help` in a dashtastic-like project with missing manifest now rebuilds before rendering help

## Notes

The Python re-export is independent of the bootstrap fix but bundled per request — it's a one-line addition to `__all__` plus the import.

### DIFF
--- a/crates/toolr-py/python/toolr/__init__.py
+++ b/crates/toolr-py/python/toolr/__init__.py
@@ -27,6 +27,7 @@ from toolr._decorators import command_group
 from toolr.build import BuildManifestError
 from toolr.build import BuildResult
 from toolr.build import build_manifest
+from toolr.sources import DispatchCommand
 from toolr.utils._imports import report_on_import_errors
 from toolr.utils._signature import ArgSection
 from toolr.utils._signature import arg
@@ -38,6 +39,7 @@ __all__ = [
     "BuildManifestError",
     "BuildResult",
     "Context",
+    "DispatchCommand",
     "__version__",
     "arg",
     "arg_section",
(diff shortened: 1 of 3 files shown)


s0undt3ch:

- #230 ◀
    - #231
        - #232

PR #231, opened 2026-05-20
argparse: skip empty files, warn on dups, complete through dispatchers

s0undt3ch:
## Summary

Three related fixes for the cases where a scan glob picks up files clap can't handle, plus completion through dispatchers.

### 1. Skip files with no `add_argument` calls

`scan_source` emitted a `ScannedCommand` for **every** file matched by the glob, regardless of whether the file actually contained any `add_argument` calls. `__init__.py` and private helpers (`_setup_*.py`) parse fine but contain zero add_argument calls — they got registered as ghost commands with empty argument lists.

When multiple `__init__.py` files were swept up by one glob (one per Django app under `web/**/management/commands/*.py`), they collided on the stem `__init__` and crashed clap at startup:

```
thread 'main' panicked at clap_builder-4.6.0/src/builder/debug_asserts.rs:353:9:
Command job: command name `__init__` is duplicated
```

Fix: in `scan_block_paths`, when a file parses successfully but produces no arguments, drop it silently. Files that fail to parse still surface as warning placeholders.

### 2. Complain loudly on intra-source duplicates

`validate_no_collisions` already errored on cross-source name clashes but silently passed intra-source duplicates through to clap. Add `dedup_intra_source` that keeps the first occurrence and emits one warning per discarded duplicate:

```
toolr: warning: argparse source "argparse:django" grafted child "sync" multiple times under "jenkins.job"; keeping the first occurrence — tighten the scan glob or rename one of the files
```

Cross-source collisions remain a hard error.

### 3. Tab completion walks through dispatchers

`toolr jenkins job <TAB>` returned nothing even though `--help` rendered the grafted children: the completion engine landed on the dispatcher Command and asked for its own positional values (it had none) instead of looking up the grafted children at `group == "jenkins.job"`.

Add a dispatcher-aware branch to `classify`:

- Dispatcher's own flag completions (`--<TAB>`, `--flag <TAB>`) still target the dispatcher's arguments.
- Positional slot 0 against a dispatcher → completing the child command name → `Slot::Command { group: <dispatcher dotted path>, prefix }`.
- Positional slot >= 1 → child chosen → recurse into that child's arg zone via the extracted `classify_leaf_args`.

Helpers `first_positional_value` and `drop_through_first_positional` walk `arg_tokens` the same way `count_positionals_consumed` does, so a parent `--cpu 200m` value pair before the child name is skipped correctly.

## Verified on dashtastic (355 grafted children)

- `toolr jenkins job delete_<TAB>` → 8 `delete_*` commands
- `toolr jenkins job --dry-run <TAB>` → still suggests children after consuming the parent flag
- `toolr jenkins job --<TAB>` → dispatcher's own flags (`--cpu`, `--dry-run`, `--job-name`, ...)
- Loose `web/**/management/commands/*.py` glob no longer panics clap
- 0 duplicate child names in the manifest

## Test plan

- [x] `scan_paths_skips_files_with_no_add_argument_calls` regression test
- [x] `dedup_intra_source_keeps_first_and_warns` + `dedup_intra_source_is_a_noop_when_no_duplicates`
- [x] 6 new dispatcher completion tests (`dispatcher_completion_lists_grafted_children`, `dispatcher_completion_filters_children_by_prefix`, `dispatcher_flag_prefix_lists_dispatcher_flags_not_children`, `dispatcher_completion_offers_children_after_parent_flag`, `dispatcher_recurses_into_child_args_once_child_is_chosen`, `dispatcher_recurses_through_parent_flag_value_pair`)
- [x] Existing `collision_is_detected` still passes (cross-source still errors)
- [x] All argparse + completion tests pass; clippy clean

## Stacks on

#230 (bootstrap rebuild on --help + DispatchCommand re-export)

### DIFF
--- a/crates/toolr-core/src/argparse/attach.rs
+++ b/crates/toolr-core/src/argparse/attach.rs
@@ -72,6 +72,47 @@ pub fn validate_no_collisions(
     Ok(())
 }
 
+/// Drop intra-source duplicates from each parent's child list, returning
+/// a warning string per duplicate that was discarded.
+///
+/// Two files matched by the same source can share a stem when an app
+/// layout puts a command of the same name in multiple Django apps, or
+/// when a glob is too broad. The historical behaviour was to silently
+/// pass them through to clap, which then panicked at startup with a
+/// "command name is duplicated" assert. Keeping the first occurrence
+/// and warning preserves the build while telling the user exactly which
+/// child name collided so they can tighten their scan glob or rename
+/// one of the files.
+///
+/// Cross-source collisions remain a hard error — they're a structural
+/// configuration problem that needs the user's resolution before any
+/// build can produce a stable manifest. See [`validate_no_collisions`].
+pub fn dedup_intra_source(
+    children_by_parent: &mut HashMap<String, Vec<Command>>,
+) -> Vec<String> {
+    let mut warnings = Vec::new();
+    for (parent, children) in children_by_parent.iter_mut() {
+        let mut seen: std::collections::HashSet<(String, String)> =
+            std::collections::HashSet::new();
+        let original = std::mem::take(children);
+        for child in original {
+            let source = child.dispatched_from.clone().unwrap_or_default();
+            let key = (child.name.clone(), source.clone());
+            if seen.contains(&key) {
+                warnings.push(format!(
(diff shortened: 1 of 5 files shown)


s0undt3ch:

- #230
    - #231 ◀
        - #232

PR #232, opened 2026-05-21
argparse: preserve literal long-flag spelling for dispatch round-trip

s0undt3ch:
## Summary

The scanner stripped `--` to produce a canonical `name` (e.g. `user-ids`) and `DispatchCommand.argv` then re-hyphenated it via `_flag_for(name)` (`name.replace("_", "-")`). That worked when the source declared a hyphenated long form (or aliased both spellings), but for Django commands that exposed **only** an underscored `--user_ids` the dispatched argv emitted `--user-ids`, which argparse on the receiving side wouldn't recognise.

dashtastic dodged the bug because every collision case in its tree happens to declare both spellings:

```python
user_selection.add_argument("--user_ids", "--user-ids", type=int, nargs="+", ...)
```

A Django command that only registered one underscored long form would have failed silently — toolr CLI looks fine, dispatch goes out with `--user-ids`, manage.py rejects it.

## Fix

Carry the source's literal long flag through the schema unchanged:

- Add `long_flag: Option<String>` to `manifest::Argument`. The argparse scanner populates it with the longest `--…` positional verbatim. `None` for positionals, native toolr commands, common-args supplied via `[tool.toolr.argparse.*]`, and third-party manifest fragments.
- Mirror the field on the wire (`ArgSchemaSpec`) and the Python schema (`ArgSchema`).
- Add `_flag_for_arg(arg)` in `_dispatch.py` — prefers `arg.long_flag` when present, falls back to `_flag_for(arg.name)` for legacy manifests / non-argparse sources. `DispatchCommand.argv` calls `_flag_for_arg` for every emitted flag.

toolr's CLI display still normalises to `--user-ids` for UX consistency; only the dispatched-to-argparse round-trip changes — and only for args whose source declared an underscored spelling.

## Test plan

- [x] New parametrised cases in `test_argv_reconstruction`:
  - `long_flag="--user_ids"` with `kind="optional"` → emits `["--user_ids", "7"]`
  - `long_flag="--user_ids"` with `kind="repeated"` → emits `["--user_ids", "3", "--user_ids", "4"]`
  - Existing `dry_run → --dry-run` case still passes (no `long_flag` → fallback hyphenation)
- [x] 393 toolr-core tests pass
- [x] 119 toolr binary tests pass
- [x] 10 sources tests pass
- [x] mypy clean
- [x] Verified end-to-end on dashtastic: `delete_companyuser` manifest now stores `long_flag` per arg; dispatch dry-run emits the literal spelling

## Stacks on

#231 (dispatcher completion + argparse skip-empty/dedup)

### DIFF
--- a/crates/toolr-core/src/argparse/scan.rs
+++ b/crates/toolr-core/src/argparse/scan.rs
@@ -243,6 +243,18 @@ fn build_argument(positionals: &[String], call: &ExprCall, warnings: &mut Vec<St
         metadata.metavar = Some(mv);
     }
 
+    // Preserve the literal long-flag spelling from the source. The
+    // scanner's canonical `name` strips the leading `--` and may pick
+    // between aliases (`add_argument("--user_ids", "--user-ids", ...)`
+    // ties on length and last-wins), so `name` alone isn't enough to
+    // round-trip the source's exact spelling. `long_flag` carries it
+    // verbatim for the dispatch path to emit unchanged.
+    let long_flag = if is_keyword_style {
+        Some(name_for_warning.clone())
+    } else {
+        None
+    };
+
     Argument {
         name,
         kind,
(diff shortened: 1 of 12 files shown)


s0undt3ch:

- #230
    - #231
        - #232 ◀

PR #240, opened 2026-05-23
fix(dispatch): emit clear error when tools venv python is missing

s0undt3ch:
## Summary

When the tools-venv Python interpreter is absent — typical case: the user blew away `~/Library/Caches/toolr/<key>/venv/`, or checked out a branch with a different repo-key hash — the dispatch path's `spawn_runner_capturing_stderr(&python, …)?` returned a bare `std::io::Error::NotFound`. `main.rs::run`'s `eprintln!("toolr: {e:#}")` then printed exactly:

```text
toolr: No such file or directory (os error 2)
```

A single line that names nothing and gives no recovery hint. Hit during \`toolr bench compare\` debugging — same opaque error appeared three times across different sessions before I located the cause.

## Fix

Pre-check `python.is_file()` before spawning, mirroring `run_introspect` (`crates/toolr-core/src/dynamic/runner.rs`, which already does this correctly for the dynamic-layer path). Emit:

```text
toolr: Python interpreter not found at <path>.
Run \`toolr project deps sync\` to materialise the tools venv.
```

Also `.with_context(...)` the spawn call so unexpected `io::Errors` (permission denied, etc.) include the python path in the chain instead of just the bare display.

## Test

`crates/toolr/tests/cli_smoke.rs::dispatch_emits_clear_error_when_venv_python_is_missing` removes the stub `python` from the existing `preflight_fixture` and asserts:

- exit code is non-zero
- stderr contains `Python interpreter not found at`
- stderr contains the `toolr project deps sync` recovery hint
- stderr is *not* just the bare `os error 2` line

## Test plan

- [ ] Manually remove `~/Library/Caches/toolr/<key>/venv/bin/python` and run any user command; verify the new message renders cleanly.
- [ ] Verify `toolr project deps sync` (the recovery hint we point at) actually rebuilds the venv from there.

### DIFF
--- a/crates/toolr/src/dispatch.rs
+++ b/crates/toolr/src/dispatch.rs
@@ -1,6 +1,7 @@
 use std::path::PathBuf;
 use std::process::ExitCode;
 
+use anyhow::Context as _;
 use clap::ArgMatches;
 
 use toolr_core::complete::{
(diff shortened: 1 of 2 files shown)


s0undt3ch:

- #240 ◀

PR #237, opened 2026-05-23
docs: fix six contradictions between docs and shipped behaviour

s0undt3ch:
## Summary

PR-1 of the [repo presentation pass](../blob/presentation-pass-1-doc-fixes/specs/2026-05-22-repo-presentation-pass-design.md). Closes six concrete contradictions where docs or doc comments misinform readers. Doc-only diff; **no runtime behaviour changes**.

## What changes

| File                                              | Change                                                                                                                                                                                              |
| ------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `docs/writing-commands/index.md`                  | Scope the deprecation admonition to the still-deprecated `parent.command_group("child", ...)` form. `@group.command` is no longer deprecated (per 2026-05-18 keep-bound-command-decorator design). |
| `docs/writing-commands/known-bugs.md`             | Closed-bugs nested-group example now uses the dotted-string form `command_group("docker.image", ...)`, not the deprecated method form.                                                            |
| `docs/internals/manifest.md` (§File shape)        | `third_party_hash` bullet now describes blake3 over `toolr-manifest.json` paths+contents — the actual implementation.                                                                              |
| `docs/internals/manifest.md` (§Dynamic layer)     | Drop the obsolete `entry_points("toolr.tools")` discovery step (removed by dispatch-manifest-freshness). Add an explicit pointer to the third-party docs.                                          |
| `docs/internals/manifest.md` (§Hashing details)   | Hashing-details bullet for `third_party_hash` now matches the §File-shape bullet and uses the same blake3 scheme description as `static_hash`.                                                     |
| `crates/toolr-core/src/dynamic/payload.rs`        | Doc comment on `DynamicPayload.groups` no longer mentions entry-point discovery.                                                                                                                    |

## Verification

- `mkdocs build --strict` succeeds after every commit (run incrementally).
- `prek run --all-files` clean — covers `rumdl`, `cargo check`, `clippy`, `codespell`, `typos`, `mypy`, `ruff`, etc.
- Diff scope is markdown + one Rust doc comment; no behavioural surface, so no `cargo test --workspace` needed.

## Test plan

- [ ] Skim each changed `docs/` page on the rendered site after merge — every cross-reference resolves.
- [ ] Confirm `docs/internals/manifest.md` reads coherently end-to-end (the two `third_party_hash` bullets are now consistent with each other and with `crates/toolr-core/src/freshness/compare.rs`).

## Stack

[...]

PR #248, opened 2026-05-25
fix(ci): pass GITHUB_TOKEN to git-cliff so [remote.github] auths

s0undt3ch:
## Summary

The release-prep workflow has been failing since the merge of `dad48f6c` (PR #237) with:

\`\`\`text
panicked at git-cliff-core/src/changelog.rs:493:18:
Could not get github metadata: HttpClientError(reqwest::Error { kind: Status(401, None),
url: "https://api.github.com/repos/s0undt3ch/ToolR/commits?per_page=100&page=0" })
\`\`\`

## Root cause

`cliff.toml` has:

\`\`\`toml
[remote.github]
owner = "s0undt3ch"
repo = "ToolR"
\`\`\`

git-cliff 2.13+ resolves that block eagerly on every run — it hits `api.github.com/repos/.../commits` to enrich the changelog body with commit metadata used by the Tera template (commit IDs, contributor PR numbers via the `remote_url` macro).

Both `git-cliff-action` invocations in `_prepare-release.yml` set `GITHUB_REPO` but **no `GITHUB_TOKEN`**, so git-cliff makes unauthenticated requests. Once it advances past the first page (or hits a rate-limited window), GitHub returns 401 and git-cliff panics — producing the trace in the issue.

The reason it surfaced after `dad48f6c`'s merge: that's when the release-prep workflow next fired against a repo state where the commit page count had grown enough to trip the unauthenticated rate ceiling.

## Fix

Add `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}` to both git-cliff-action `env:` blocks. The default workflow token has the read-only repo scope git-cliff needs; no new permissions, no new secrets, no behavioural change on success.

Inline comments in both steps explain *why* the token is needed (prevents the next editor stripping it back out as unused-looking env).

## Verification

- `prek run actionlint --files .github/workflows/_prepare-release.yml`: clean.
- Confirm fix by triggering the release workflow against an unreleased state once this merges.

## Test plan

- [ ] Dispatch `release.yml` against an unreleased state (or push a tag locally to simulate) and confirm the `Generate changelog for GH release` step succeeds.
- [ ] `Update CHANGELOG.md` step in the same workflow also succeeds (same fix applied).

### DIFF
--- a/.github/workflows/_prepare-release.yml
+++ b/.github/workflows/_prepare-release.yml
@@ -180,6 +180,14 @@ jobs:
           # $GITHUB_ENV write; cliff.toml reads it via Tera get_env.
           OUTPUT: .release-changes.md
           GITHUB_REPO: ${{ github.repository }}
+          # cliff.toml's `[remote.github]` block makes git-cliff hit
+          # the GitHub API for commit/PR metadata used by the body
+          # template. Without an explicit token it falls through to
+          # unauthenticated requests and the API returns 401 on
+          # private endpoints / once rate-limited. The default
+          # workflow token is sufficient and has read-only repo
+          # scope.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Prepare GH Release changelog
         id: prepare-release-notes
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Closing: the original 401s were a 2026-05-23 GitHub-side flake, not a real auth issue. The same workflow ran successfully against main today (run 26328110637, job 77669844900) without this change. Re-open if the fluke recurs.

PR #252, opened 2026-05-25
test(command): wall-clock-bound tokio command tests, fix #247 workspace hang

### LINKED ISSUE #247: test: `test_no_output_timeout` flakes under parallel workspace load
PR #243 fixed the bare-`python` deadlock variant of this test by introducing a `python3`-first resolution helper. The python-resolution fix is correct, but the test **still hangs** when run under `cargo test --workspace` (parallel load), even when `python3` is callable. It passes consistently when run in isolation.

## Reproduce

On a system with `python3` on PATH and `cargo test -p toolr-core --lib tokio_tests::test_no_output_timeout` passing in <1s:

\`\`\`sh
cargo test --workspace
\`\`\`

Cargo will eventually print:

\`\`\`text
test command::tests::test_suite::tokio_tests::test_no_output_timeout has been running for over 60 seconds
\`\`\`

and stay there until interrupted. The test was observed running ≥3 minutes during the 2026-05-23 cleanup session before being killed.

## What we know

1. **Isolation passes.** `cargo test -p toolr-core --lib tokio_tests::` runs the same four tokio tests (including `test_no_output_timeout`) and completes in ~0.6s with all four green.
2. **Bare-`python` spawn-deadlock is fixed.** PR #243's `resolve_python_for_test()` + `require_python!` macro ensures the python interpreter resolves before the test body runs.
3. **The watchdog setup works in principle.** `no_output_timeout_secs: 0.1` with a Python child that flushes once and then `time.sleep(10)` is the right shape — the no-output watchdog at `crates/toolr-core/src/command/` is supposed to fire after 100ms of silence and kill the child.
4. **Killed processes leave zombies.** During investigation, multiple zombie `toolr_core-<hash>` test-runner processes were observed (one per hung run). They had no python child but were stuck inside their own thread. Suggests the hang is *inside* the test (probably the `handle.join()` after the pipe read), not the python child.

## Hypotheses, not yet verified

- **Tokio scheduling under contention.** When `cargo test --workspace` runs many crates' tests in parallel, the tokio runtime in `test_no_output_timeout` gets fewer time slices. If the watchdog task starves long enough that the no-output threshold (0.1s) lapses against wall-clock but the watchdog future isn't polled, the kill might never fire. Once python's 10s sleep completes naturally, the test then takes ≥10s — but observed hangs were much longer than 10s, so this can't be the whole story.
- **Pipe write-end ownership.** The test process holds `stdout_write` alive (in addition to passing it to the child via `sys_stdout_fd`). The pipe-write end isn't closed until *after* `handle.join()`. If `run_command_internal` somewhere internally tries to drain the pipe before returning, it may block on a writer that the test process is still holding open. This would explain the indefinite hang (not just 10s).
- **`read_pipe` ordering.** The test does \`read_pipe(stdout_read, ...)\` *before* \`handle.join()\`. If the watchdog kills python but the test's read has already returned and is now sitting at `join()`, but the thread is wedged on its own pipe close, we get a deadlock loop.

## Suggested next steps

1. **Add a wall-clock guard around `handle.join()`** so the test fails fast instead of hanging forever. \`std::sync::mpsc::Receiver::recv_timeout\` against a channel the thread signals on completion is the standard pattern. 5s should be ample (the test should complete in ~0.3s).
2. **Close test-side pipe write FDs before the read.** After spawning the thread, the test process no longer needs the writer end; closing it allows EOF to propagate naturally when the child exits.
3. **Audit the no-output watchdog implementation** in `crates/toolr-core/src/command/` for the specific case where the child writes once before going silent. Verify the watchdog future polls within the deadline under contention.
4. **Until fixed: `#[ignore]` the test** with a doc comment pointing here. The behaviour it exercises is real but the harness deadlocks more than it observes the no-output path.

## Context

Filed during the 2026-05-23 codebase-audit session (see [`specs/2026-05-23-codebase-audit.md`](../blob/audit-deep-dive/specs/2026-05-23-codebase-audit.md) Plan 1). The audit's Plan 1 description estimated this as "S, safe, independent" — that turned out to cover only the spawn-deadlock half. This issue tracks the other half.

s0undt3ch:
Closes #247.

## Summary

`test_no_output_timeout` deadlocked under `cargo test --workspace` even though it passed in
isolation. PR #243 covered the spawn-failure half (`python3` resolution); this PR closes the
other half (pipe-read deadlock under parallel-test contention) and adds a wall-clock guard so
no future regression in this code path can hang the workspace.

## Root cause

The test wired its own pipe through `sys_stdout_fd`, then read it *before* joining the worker:

```rust
let handle = thread::spawn(move || run_command_internal(config));

// Read first — BEFORE join. If nothing is ever written, this blocks forever.
read_pipe(stdout_read, &mut stdout_buffer);

let result = handle.join().expect("Thread panicked");
```

Under workspace contention the no-output watchdog could fire and kill python *before* the
in-process stdout forwarder had read anything to write to the test-owned pipe:

1. python is slow to start under load → no bytes on `child.stdout` yet.
2. Watchdog trips (`no_output_timeout_secs: 0.1` is very tight) and kills python.
3. Forwarder thread reads EOF and exits without ever writing to `sys_stdout_fd`.
4. Test is still parked in `read_pipe(stdout_read, …)`. The write end (`stdout_write`) is still
   open on the test process, so the kernel never delivers EOF. The read blocks indefinitely.
5. `handle.join()` is never reached.

## Fix

All changes are in `crates/toolr-core/src/command/command_test.rs`.

**1. New `bounded_run_command` helper in `tokio_tests`.**
Runs `run_command_internal` via `tokio::task::spawn_blocking` and wraps the join in
`tokio::time::timeout`. Any future regression that wedges the call fails the test fast with a
clear ``run_command_internal did not return within 5s`` message instead of hanging the workspace.

**2. `test_no_output_timeout` simplified.**
- Drop the test-owned `sys_stdout_fd` / `sys_stderr_fd` pipes entirely. The watchdog updates
  `last_output` from the parent-side read against `child.stdout`, so a sys-fd consumer is not
  needed to exercise the no-output path. This removes the deadlock surface completely.
- Bump `no_output_timeout_secs` from `0.1` to `0.5`. Under scheduler jitter the old value could
  trip *before* python had a chance to write anything; `0.5s` lets the test actually exercise
  the "wrote, then went silent" path.
- Call wrapped in `bounded_run_command(5s)` for belt-and-braces.

**3. `test_timeout_exception` hardened the same way.**
Same shape (direct `run_command_internal` call, very tight timeout), so the same guard applies.
`timeout_secs` `0.1` → `0.5`, call wrapped in `bounded_run_command(5s)`.

## Verification

```text
cargo test --workspace          # 3 back-to-back runs, all green
test result: ok. 419 passed; 0 failed; … finished in 3.04s
test result: ok. 419 passed; 0 failed; … finished in ~2.9s
test result: ok. 419 passed; 0 failed; … finished in ~2.9s

cargo clippy --workspace --all-targets -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s)

prek run --files crates/toolr-core/src/command/command_test.rs
# all hooks pass
```

Before this PR the workspace run hung indefinitely on `test_no_output_timeout` and had to be
killed manually (observed ≥3 minutes during the 2026-05-23 audit session).

### DIFF
--- a/crates/toolr-core/src/command/command_test.rs
+++ b/crates/toolr-core/src/command/command_test.rs
@@ -433,6 +433,31 @@ mod test_suite {
         use super::*;
         use tokio::runtime::Runtime;
 
+        /// Run `run_command_internal` on a blocking worker thread, bounded by a
+        /// wall-clock timeout. Returns the inner result on completion, or an
+        /// `Err` describing the wall-clock breach.
+        ///
+        /// The bound exists so a future regression that wedges
+        /// `run_command_internal` (a stuck forwarder thread, a missed kill, etc.)
+        /// fails the test fast instead of hanging the entire workspace test run.
+        /// See issue #247 for the original `cargo test --workspace` hang this
+        /// guards against.
+        async fn bounded_run_command(
+            config: CommandConfig,
+            timeout: Duration,
+        ) -> Result<std::result::Result<i32, Box<dyn std::error::Error + Send + Sync>>> {
+            let join_handle =
+                tokio::task::spawn_blocking(move || run_command_internal(config));
+            match tokio::time::timeout(timeout, join_handle).await {
+                Err(_) => Err(anyhow!(
+                    "run_command_internal did not return within {:?}",
+                    timeout
+                )),
+                Ok(Err(e)) => Err(anyhow!("worker task failed: {e}")),
+                Ok(Ok(inner)) => Ok(inner),
+            }
+        }
+
         // Cross-platform pipe creation
         #[cfg(unix)]
         fn create_pipe() -> Result<(i32, i32)> {
@@ -637,12 +662,13 @@ mod test_suite {
                         "-c".to_string(),
                         "import time; time.sleep(10)".to_string(),
                     ],
-                    timeout_secs: Some(0.1), // Very short timeout
+                    timeout_secs: Some(0.5), // Short timeout; 0.5s tolerates workspace contention
                     ..Default::default()
                 };
 
-                // Run the command and expect failure
-                let result = run_command_internal(config);
+                // Wall-clock-bounded so the workspace test run can't hang if the
+                // internal kill+wait path ever wedges. See issue #247.
+                let result = bounded_run_command(config, Duration::from_secs(5)).await?;
 
                 // Should fail with timeout
                 assert!(result.is_err());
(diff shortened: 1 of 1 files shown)


s0undt3ch:

- #252 ◀

PR #253, opened 2026-05-25
feat(dispatch): inherit child stderr to preserve TTY; move missing-dep hint into the runner

s0undt3ch:
Rich's terminal detection in user commands was failing because every
toolr dispatch piped the runner's stderr through an in-process pumper
that captured the whole stream so we could post-mortem-rewrite an
``ImportError`` traceback into a styled "run `toolr project deps sync`"
report. Side effect: ``sys.stderr.isatty()`` was always False in the
child, Rich emitted no ANSI/OSC 8 on stderr, and stderr only landed
in the terminal as a burst after the child exited.

This change drops the post-mortem capture and re-issues the same hint
from inside the runner — which is our own code and can emit styled
output to its own inherited stderr without breaking TTY semantics.
Pre-flight diagnostics (declared imports) remain in Rust and now cover
extension modules and PEP 420 namespace packages so they aren't
false-positives for `.so`/`.pyd`/namespace-shaped deps.

## Rust side

- `crates/toolr/src/dispatch.rs` — use `spawn_runner` (inherit stdin/
  stdout/stderr) instead of `spawn_runner_capturing_stderr`. Delete
  the `intercept_import_error` post-mortem branch. The runner's
  stderr now goes straight to the terminal in real time.
- `crates/toolr-core/src/execute/spawn.rs` — delete `StderrCapture`
  and `spawn_runner_capturing_stderr` (and their pumper thread); drop
  the no-longer-needed `io::Read` / `sync::{Arc,Mutex}` / `thread`
  imports.
- `crates/toolr-core/src/execute/mod.rs` — drop the deleted symbols
  from the re-exports.
- `crates/toolr-core/src/deps_check/mod.rs` — delete the `post_mortem`
  module and its re-exports. Rewrite the module-level doc to be
  accurate about pre-flight scope and to call out that transitive
  imports are now handled runner-side.
- `crates/toolr-core/src/deps_check/post_mortem.rs` — removed.
- `crates/toolr-core/src/deps_check/probe.rs` — extend `ProbeOutcome`
  with `Extension(PathBuf)` and `NamespacePackage(PathBuf)`. Teach
  `probe_module` to recognise bare `<top>.so` / `<top>.pyd`,
  ABI-tagged extensions (`<top>.abi3.so`, `<top>.cpython-313-darwin.so`,
  …) via a directory scan, and PEP 420 namespace packages (bare
  `<top>/` with no `__init__.py`). Drop the stale "Falls through to
  post-mortem" claim — the pre-flight never reached the post-mortem
  in the first place.
- `crates/toolr-core/src/deps_check/tests.rs` — drop the post-mortem
  unit tests; extend `fake_venv` with `ext_so`/`ext_pyd`/`ext_abi3`/
  `ext_cpython`/`namespace` shapes; add eight new probe tests for
  the new variants and the prefix-collision guard.
- `crates/toolr/tests/cli_smoke.rs` — delete the post-mortem-specific
  integration tests (`post_mortem_rewrites_import_error_output`,
  `pre_flight_and_post_mortem_split_against_one_project`) and the
  `two_command_fixture` helper that only existed to feed them; add
  a regression test that the runner's stderr passes through Rust
  unaltered.

## Python side (runner)

- `crates/toolr-py/python/toolr/_runner.py` — add
  `_print_missing_dep_hint(exc, stream)` that emits the same
  "import `X` failed at runtime — run `toolr project deps sync`"
  text the Rust post-mortem used to produce. Wire it into `run()` in
  two places: the existing `SpecError` arm now calls it when the
  underlying `__cause__` is an `ImportError` (covers top-level and
  transitive import failures wrapped by `_import_target`), and a
  new `except ImportError` arm catches imports raised from inside
  the command function body. Fall back to a generic "this module"
  placeholder when the `ImportError.name` attribute is `None` (bare
  ``from pkg import thing`` form).
- `tests/runner/test_runner_coverage.py` — three new tests covering
  function-body, transitive-via-SpecError, and bare-ImportError
  branches; tighten the existing unhandled-exception test to assert
  the hint does *not* fire for non-import errors.

## Design doc

- `specs/2026-05-25-preserve-child-tty-design.md` → archived under
  `specs/archive/2026/`.

## Verification

```text
cargo test --workspace
# all binaries pass, ~620 tests total, finished in ~7s

cargo clippy --workspace --all-targets -- -D warnings
# clean

prek run --files <changed files>
# all hooks pass

uv run pytest tests/runner/ -q
# 74 passed
```

Manual: `toolr` commands that print Rich markup with `[link=URL]…[/link]`
to stderr now render clickable OSC 8 hyperlinks on TTY terminals
instead of stripped bare URLs. A command that imports a missing
module still produces the styled "run `toolr project deps sync`"
guidance — emitted by the runner this time.

Closes the design at specs/archive/2026/2026-05-25-preserve-child-tty-design.md
(Option A).


s0undt3ch:

- #253 ◀

PR #266, opened 2026-05-25
feat(argparse): `T | None` without default → zero-or-one positional

s0undt3ch:
## Summary

Make `T | None` (without a default) a proper zero-or-one positional argument — the type-driven replacement that the `arg(nargs="?")` deprecation message already pointed users at. Two runtime gaps closed:

- **Python runner** (`_coerce_args`): clap was already accepting the omission (the resolved `Optional(_)` type flips `required→false` on the positional), but the runner was leaving the parameter out of the call entirely, so the function blew up with `TypeError: missing 1 required positional argument`. The runner now injects `None` for any absent parameter whose annotation is `T | None` and which has no default.
- **Python signature parser** (legacy argparse path): didn't set `nargs="?"` on bare-positional `T | None`, so argparse-driven code paths still treated them as required.

## Semantics

| Annotation | Result |
|---|---|
| `name: str` | required positional |
| `name: str \| None` | **zero-or-one positional** (this PR) |
| `name: str \| None = None` | `--flag` (unchanged) |

```sh
toolr version bump              # new_version → None
toolr version bump 0.20.0       # new_version → "0.20.0"
```

## Structural rules (enforced at manifest-build time)

To prevent ambiguity at the parser layer, three rules are enforced — both in the rust static manifest builder (`BuildError::InvalidPositionalArity`) and the python legacy argparse path (`SignatureError`):

1. **At most one** zero-or-one positional per command. Two trailing optionals can't be disambiguated.
2. The zero-or-one slot must come **after all required positionals**. Fixed-arity `tuple[T1, T2, …]` positionals may freely precede it — they're deterministic.
3. `T | None` and `*args: T` **cannot coexist** — both consume the trailing slot.

Example error:
```
toolr: invalid positional arity (1):
  - tools.foo::two-optionals: two zero-or-one positionals `a` and `b`. Only one
    `T | None` positional is allowed per command; give one of them a default to
    make it a `--flag`.
```

## Also in this PR

Drops the deprecated `Annotated[str | None, arg(nargs="?")]` wrapper on `tools/version.py::bump` now that the type-driven form is sufficient.

## Test plan

- [x] `cargo test --workspace` — 426 + smaller test binaries, all green
- [x] `uv run pytest` — 389 passed, 8 skipped (distribution tests, no wheels in workspace)
- [x] `uv run prek run --files <changed>` — clean
[...]

PR #268, opened 2026-05-26
fix(version): drop spurious `= None` default on bump's positional

s0undt3ch:
## Summary

Drop the spurious `= None` default on `tools/version.py::bump`'s `new_version` parameter — a carryover from #266 where the file was accidentally restored from a working-tree stash with `= None` still attached.

## Why this matters

With `new_version: str | None = None`, the parameter classifies as a **keyword flag** (`--new-version VALUE`) under the new 0.20.0+ semantics — and the release-prep workflow calls:

```yaml
- name: Bump Version
  run: toolr version bump --write --check-existing-tag ${{ inputs.release-version }}
```

passing the version *positionally*. That would error with `unexpected argument '0.20.0' found` and break the release.

## The right shape

`new_version: str | None` (no default) accepts the value positionally under both runtimes:
- **0.11.1** (used during release prep): treated as a required positional. The workflow always passes the value, so this works.
- **0.20.0+** (post-release, after #266 lands): zero-or-one positional. `toolr version bump` (no arg) takes the dev-version path; `toolr version bump 0.20.0` sets the explicit version.

Verified locally against the freshly-built toolr binary:

```
$ toolr version bump
0.11.2-dev538

$ toolr version bump 0.20.0
0.20.0

$ toolr version bump --check-existing-tag 0.20.0
0.20.0
```

### DIFF
--- a/tools/version.py
+++ b/tools/version.py
@@ -100,7 +100,7 @@ def current(ctx: Context) -> None:
 @group.command
 def bump(
     ctx: Context,
-    new_version: str | None = None,
+    new_version: str | None,
     check_existing_tag: bool = False,
     write: bool = False,
 ) -> None:


s0undt3ch:

- #268 ◀

PR #270, opened 2026-05-26
ci(build): drop `cross`, build musl natively, run full matrix on main

s0undt3ch:
## Summary

Three coupled CI changes that share the same answer — *"treat main pushes like releases, give PRs an opt-in escape hatch, and put the decision logic next to the matrix data."*

1. **Drop `cross` entirely.** Both `*-linux-musl` triples already run on architecture-matched runners, so cross was paying for itself with nothing. It also just broke outright — see [this run](https://github.com/s0undt3ch/ToolR/actions/runs/26437125416/job/77830040609#step:11:16): cross 0.2.5 unconditionally runs `rustup toolchain add stable-x86_64-unknown-linux-gnu`, which modern rustup refuses on the aarch64 runner.
2. **Push-to-main + opt-in label run the full release matrix.** PRs stay fast (3 native triples + 1 CPython ABI for the binary wheel); main pushes and any PR carrying the `full-build` label exercise every triple and every CPython ABI that `release.yml` would build — so musl + cross-arch regressions get caught before tag-time.
3. **The matrix-selection logic lives in `tools/ci.py`, not in inline YAML.** A new `_select_workflow_mode()` returns `(mode, reason)`, and `generate-build-matrix` writes the reason as a one-line `_italics_` blurb under the `## Build matrix` job-summary heading. No more hunting through workflow files to figure out *why* a run picked what it picked.

Supersedes #269 (the `--locked` workaround on `cargo install cross`) — there's no cross install anymore.

## Why was cross there at all?

The historical intent: \"cross handles the musl libc + linker setup so we don't have to.\" That was reasonable when the project was younger and the matrix didn't have arm-native GitHub runners.

It's no longer true:
- We already pick architecture-native runners (`ubuntu-latest` for x86_64-musl, `ubuntu-24.04-arm` for aarch64-musl), so there's nothing to *cross-compile* — only a libc to swap.
- `rustup target add <triple>-musl` + `apt-get install musl-tools` (provides `musl-gcc`) is a 2-line native build.
- `cross 0.2.5` has been the only published release since 2023-02-04 (project is alive on `main`, but unreleased), and its frozen `Cargo.lock` keeps pinning transitives that later get yanked (e.g. `dunce v1.0.2`). Every CI run was emitting `--locked` warnings.

## How the matrix is selected (first match wins)

| Event | Picked | Reason rendered to job summary |
|---|---|---|
| `push` to `refs/heads/main` | `release` | _push to `main` — full matrix mirrors the release surface._ |
| `pull_request` with `full-build` label | `release` | _PR carries the `full-build` label — running the full release matrix on opt-in._ |
| `pull_request` without that label | `ci` | _PR (no `full-build` label) — native-triple subset to keep PR builds snappy. Apply the `full-build` label and re-run to opt into the full release matrix._ |
| explicit `--workflow release` (release.yml) | `release` | _caller forced `--workflow release`._ |
| everything else (push to non-main, tag, manual dispatch) | `ci` | _event=`<name>`, ref=`<ref>` — default to the minimal CI matrix._ |

`ci.yml`'s `pull_request` trigger gained `labeled` + `unlabeled` so applying or removing the `full-build` label re-runs the workflow immediately — no dummy commit needed.

## Changes

**`tools/ci.py`**
- Drop the `cross: True/False` field from every `_BINARY_ARCHIVE_TRIPLES` entry.
- New `FULL_BUILD_LABEL = \"full-build\"` constant.
- New `_pr_labels()` reads `$GITHUB_EVENT_PATH` for the PR's labels.
- New `_select_workflow_mode()` returns `(mode, reason)` from the event context.
- `generate_build_matrix(workflow: Literal[\"ci\", \"release\"] | None = None)` — `None` triggers auto-detect, an explicit value bypasses it. Writes a `_reason_` italics line to the job-summary block.
- Step-summary table for binary archives loses its now-irrelevant \"Build mode\" column.

**`.github/workflows/_build-binary-archive.yml`**
- Delete \"Install cross\" + \"Build binary (cross)\" steps.
- Delete the `if: ${{ !matrix.target.cross }}` gating — there's only one path now.
- Add `Install musl-tools`, conditional on `contains(matrix.target.triple, '-linux-musl')`.
- Single `Build binary` step: `cargo build --release -p toolr --bin toolr --target <triple>`.

**`.github/workflows/ci.yml`**
- `pull_request` trigger now lists `types: [opened, synchronize, reopened, labeled, unlabeled]`.
- `generate-build-matrix` invocation drops the env var + inline conditional; just runs `toolr ci generate-build-matrix` and lets `tools/ci.py` decide.

## Test plan

- [x] `cargo build -p toolr` + `toolr project manifest rebuild` clean.
- [x] All five selector paths exercised locally:
  - default (no GH env) → `ci` with the unset-event reason
  - `push` to `refs/heads/main` → `release` with the main-push reason
  - PR with `full-build` label → `release` with the opt-in reason
  - PR without the label → `ci` with the apply-label-to-opt-in reason
  - explicit `--workflow release` → `release` with the forced-by-caller reason
- The real test is this PR itself: the CI matrix should pick `ci` (no `full-build` label). Apply the label to flip it to `release` on demand.

### DIFF
--- a/.github/actions/throttle/action.yml
+++ b/.github/actions/throttle/action.yml
@@ -1,0 +1,17 @@
+name: Throttle Build
+description: Throttle the build to avoid thundering herd effects.
+
+inputs:
+  max-sleep-seconds:
+    description: The maximum number of seconds to sleep.
+    required: true
+    default: '5'
+
+runs:
+  using: composite
+
+  steps:
+    - name: Configure Git
+      shell: bash
+      run: |
+        t=$(python3 -c 'import random, sys; sys.stdout.write(str(random.randint(1, ${{ inputs.max-sleep-seconds }})))'); echo "Sleeping $t seconds"; sleep "$t"
--- a/.github/workflows/_build-binary-archive.yml
+++ b/.github/workflows/_build-binary-archive.yml
@@ -57,6 +57,8 @@ jobs:
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
+      - uses: ./.github/actions/throttle
+
       - name: Download Source Tarball Artifact
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
--- a/.github/workflows/_build-binary-archive.yml
+++ b/.github/workflows/_build-binary-archive.yml
@@ -22,7 +22,7 @@ on:
       targets:
         required: true
         type: string
-        description: 'JSON-encoded array of {triple, runner, cross, archive} objects describing which standalone binaries to build. Produced by `toolr ci generate-build-matrix --workflow {ci,release}` so the matrix stays centralized in tools/ci.py.'
+        description: 'JSON-encoded array of {triple, runner, archive} objects describing which standalone binaries to build. Produced by `toolr ci generate-build-matrix --workflow {ci,release}` so the matrix stays centralized in tools/ci.py.'
 
 permissions:
   contents: read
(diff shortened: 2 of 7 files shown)


s0undt3ch:

- #270 ◀

PR #272, opened 2026-05-26
feat(arg): validate aliases / conflicts_with / requires kwargs

s0undt3ch:
## What this fixes

`arg(conflicts_with="other")` was silently accepted by the Python helper (msgspec doesn't validate plain construction) and then silently dropped by the Rust AST parser (which only matched list literals) — leaving the user with a working-looking command whose mutex never fired. Hit in a real downstream consumer earlier today.

### `arg()` kwargs now validate at the boundary

| Kwarg            | Accepted shapes                                | Rejected                       | Why                                                  |
| ---------------- | ---------------------------------------------- | ------------------------------ | ---------------------------------------------------- |
| `aliases`        | `list` / `tuple`                               | set, frozenset, bare str       | Order drives clap's `Arg::short` assignment          |
| `conflicts_with` | `list` / `tuple` / `set` / `frozenset`         | bare str, dict, generators     | Semantically set-like                                |
| `requires`       | `list` / `tuple` / `set` / `frozenset`         | bare str, dict, generators     | Semantically set-like                                |

All accepted shapes coerce to `list[str]` internally. Rejection raises `TypeError` with the correct form to use, and pinpoints the bad element index when a member isn't a string.

Two type aliases capture the split:

```python
_StrSequence:   TypeAlias = list[str] | tuple[str, ...]
_StrCollection: TypeAlias = list[str] | tuple[str, ...] | set[str] | frozenset[str]
```

### Rust AST parser made symmetric

`literal_str_list` now matches `Expr::List`, `Expr::Tuple`, and `Expr::Set` — the same source-level literals the Python helper accepts at runtime. Anything more dynamic (name references, function calls) is still runtime-only and won't appear in the static manifest.

## Tests

| File                                                          | What's new                                                                                                                                                                                            |
| ------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `tests/utils/signature/test_argument_annotation.py`           | 10 cases: bare-string rejection on all three kwargs; `aliases` accepting list+tuple / rejecting sets; `conflicts_with` / `requires` accepting list/tuple/set/frozenset and rejecting dicts/non-strings |
| `crates/toolr-core/src/parser/types/mod.rs`                   | 3 parser cases: tuple literal, set literal, defensive bare-string case                                                                                                                                |
| `crates/toolr/tests/conflicts_with_dispatch.rs` *(new file)* | 3 end-to-end integration tests: clap rejects `--a --b` with exit 2, single-flag passes the gate, `--help` still renders. **This was the explicit coverage gap that let the downstream bug through.** |

## Verified

- [x] `uv run pytest tests/utils/signature/` — 73 pass
- [x] `cargo test -p toolr-core --lib parser::types::tests` — 27 pass
- [x] `cargo test -p toolr --test conflicts_with_dispatch` — 3 pass
- [x] `prek` on changed files — ruff / mypy / clippy / typos all clean

### DIFF
--- a/crates/toolr-core/src/parser/types/literals.rs
+++ b/crates/toolr-core/src/parser/types/literals.rs
@@ -11,14 +11,24 @@ pub(super) fn literal_str(expr: &Expr) -> Option<String> {
     }
 }
 
+/// Extract a string sequence from a list, tuple, or set literal of
+/// string literals. Mirrors the Python `arg()` helper, which accepts
+/// any non-string `Iterable[str]` — these three are the
+/// statically-recognisable literal forms.
+///
+/// Returns `None` for any other expression shape (name references,
+/// function calls, mixed-type literals). Callers should surface that
+/// as a manifest-build warning rather than silently dropping the
+/// kwarg.
 pub(super) fn literal_str_list(expr: &Expr) -> Option<Vec<String>> {
-    let Expr::List(list) = expr else { return None };
-    let out: Vec<String> = list.elts.iter().filter_map(literal_str).collect();
-    if out.len() == list.elts.len() {
-        Some(out)
-    } else {
-        None
-    }
+    let elts: &[Expr] = match expr {
+        Expr::List(list) => &list.elts,
+        Expr::Tuple(tup) => &tup.elts,
+        Expr::Set(set) => &set.elts,
+        _ => return None,
+    };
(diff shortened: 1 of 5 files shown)


s0undt3ch:

- #272 ◀

PR #278, opened 2026-05-28
mise: drop in-tree asdf plugin, install via aqua backend

s0undt3ch:
## Summary

Two coordinated changes that together simplify the mise install story and shorten CI:

1. **Drop the in-tree asdf-style mise plugin** in favour of mise's built-in aqua backend, now that toolr is published in the [aqua registry](https://github.com/aquaproj/aqua-registry/tree/main/pkgs/s0undt3ch/ToolR) via [aquaproj/aqua-registry#54458](https://github.com/aquaproj/aqua-registry/pull/54458).
2. **Take `tools/` out of the uv workspace** so its venv resolves `toolr-py` from PyPI (matching what external Setup ToolR consumers experience) and no longer triggers a maturin source build on every CI run.

Each change stands on its own; they're bundled because they share the "stop maintaining bespoke install plumbing we don't need" thesis.

## 1 — mise plugin → aqua backend

End users now install with:

```sh
mise use aqua:s0undt3ch/ToolR@latest
```

No plugin to register, no `git::<url>//<subdir>` syntax to maintain, no in-tree asdf-protocol hooks. The aqua-registry entry pulls the same signed GitHub release archive the in-tree plugin used to fetch, so the installed binary is identical.

**Deletes:**

- `installation/mise/` (download/install/list-all hooks + README) — replaced by mise's built-in aqua backend.

**Rewrites:**

- `docs/installation/mise.md` — aqua-backed flow as "Pin per project (recommended)", `--global` as an opt-in for machine-wide installs, troubleshooting, and a migration block.
- `README.md`, `docs/quickstart.md`, `docs/installation/index.md` — mise sections become a one-liner `mise use aqua:s0undt3ch/ToolR@latest` (project-scoped; `--global` is documented in the mise-specific page).
- `.github/workflows/install-smoke.yml` — removes the local-staging `smoke-mise-plugin` job and replaces the broken `smoke-mise-plugin-url` job with `smoke-mise-aqua` that exercises the user-visible install path via `mise install` + `mise exec` against the published aqua-registry entry. Path-trigger globs drop `installation/mise/**`.
- `UNRELEASED.md` — replaces the two outdated 0.20.0 plugin notes with a single Removed entry + migration commands.

### Migration for existing users

```sh
mise plugin uninstall toolr
mise use aqua:s0undt3ch/ToolR@latest
```

(Add `--global` for machine-wide; project-scoped is the default the docs now show.)

### Timing note

The aqua-registry merge landed after the latest `v4.517.0` release tag (2026-05-26). Users who try `mise use aqua:s0undt3ch/ToolR` before the next aqua-registry release ships will see `no aqua-registry found for s0undt3ch/ToolR` from mise. Once `v4.518.0` (or later) is cut, the install works end-to-end.

## 2 — `tools/` leaves the uv workspace

The toolr repo's `tools/` directory was a uv workspace member of the root, sharing the root `uv.lock`. Because `[tool.uv.workspace] members` included `crates/toolr-py`, every `uv sync` against the tools venv resolved `toolr-py` to `editable = "crates/toolr-py"` — which made maturin build the pyo3 extension from source on every run, including in CI. The four in-repo `Setup ToolR` callers (`_prepare-release.yml`, `ci.yml`, `sync-rolling-tags.yml`, `release.yml`) all paid this cost.

The fix is architectural: take `tools/` out of the uv workspace and give it its own lockfile. `toolr-py` then resolves from PyPI exactly the way it does for external Setup ToolR consumers.

**Changes:**

- Root `pyproject.toml`: `tools` removed from `[tool.uv.workspace] members` (alongside the previously-commented-out `crates/toolr-py`). Root workspace now contains only the toolr crates plus `examples/plugin-package`.
- `tools/pyproject.toml`: standalone uv project. `[tool.uv.sources]` block stays commented out; the comment now documents a path-based editable override (`toolr-py = { path = "../crates/toolr-py", editable = true }`) for the rare local-iteration workflow.
- `uv.lock`: regenerated. `toolr-py` resolves to the PyPI registry. Root `dev` / `docs` / `tools` dependency-groups pull the prebuilt wheel.
- `tools/uv.lock` (NEW): standalone lockfile, 8 packages from PyPI. No maturin build invoked by `uv sync --project tools`.

**Why uv requires this shape:** validated empirically (see comment thread). uv enforces `[tool.uv.sources]` for any workspace-member name in `dependencies`, and the workspace source applies unconditionally regardless of `extra =` qualifiers. The "default = PyPI, opt-in workspace via extras" pattern doesn't work in uv 0.11. Taking `tools/` out of the workspace is the only configuration that actually achieves "PyPI by default".

### Effect

- **External Setup ToolR consumers:** unchanged. Their `tools/pyproject.toml` always resolved `toolr-py` from PyPI.
- **In-repo CI:** `uv sync` against `tools/.venv` now installs the prebuilt PyPI wheel including the precompiled `_rust_utils.cpython-*.so`. No maturin compile. Faster every job.
- **Contributors:** the common case (editing `tools/*.py` without touching `toolr-py`) is unaffected — `uv sync --project tools` is now a plain PyPI install. Iterating on `crates/toolr-py/` against `tools/*.py` requires uncommenting the path source in `tools/pyproject.toml` and re-syncing. The dogfooding gap is acceptable because public-surface changes to toolr-py are caught by `cargo test --workspace` and the skill-refs `--check` gate.

## Test plan

- [x] `mkdocs build --strict` clean.
- [x] `prek run --all-files` all hooks pass.
- [x] `uv lock` at root regenerates with `toolr-py` from PyPI registry.
- [x] `uv lock` inside `tools/` writes `tools/uv.lock` with toolr-py from PyPI.
- [x] `uv sync --project tools` produces `.venv` with the prebuilt `_rust_utils.cpython-314-darwin.so` (no maturin build).
- [x] `toolr version current` dispatches to `tools/version.py` and prints the workspace version.
- [ ] After aqua-registry next release: smoke `mise use aqua:s0undt3ch/ToolR@latest` locally.
- [ ] After merge: confirm `smoke-mise-aqua` CI job passes against the published aqua-registry entry.
- [ ] After merge: confirm in-repo CI workflows no longer build toolr-py from source during Setup ToolR.


s0undt3ch:

- #278 ◀

PR #279, opened 2026-05-28
ci: drop self action-version pin and auto-sync rolling tags on release

s0undt3ch:
## Summary

Stop having the `sync-rolling-tags` workflow rewrite our own
`uses: s0undt3ch/ToolR@…` references on every release, and instead run
the workflow automatically when a release is published or deleted so
the major/minor rolling tags (`vX`, `vX.Y`) stay current without
manual `workflow_dispatch` runs.

## Why

- `release.yml` already updates the in-repo action references during
  release preparation, so `_update_action_version` inside
  `sync_rolling_tags` was redundant and a frequent source of
  empty/no-op commits.
- The rolling tags themselves should move automatically whenever the
  release set changes — both on publish (new latest patch) and on
  delete (re-point or remove a rolling tag).

## What changed

- **`tools/ci.py`** — remove the `_update_action_version` helper and
  its call site in `sync_rolling_tags`. The command now only manages
  the rolling major/minor tags.
- **`.github/workflows/sync-rolling-tags.yml`**
  - Add `release: { types: [published, deleted] }` to the triggers.
  - Gate the admin permission check on `workflow_dispatch` only, since
    on a `release` event the triggering actor is the GitHub App rather
    than the human admin and would otherwise fail the check.
- **`.github/workflows/release.yml`** — pass the GitHub App token to
  `softprops/action-gh-release`. Releases created with the default
  `GITHUB_TOKEN` do not fire downstream workflow events, so without
  this the new `release: published` trigger would never actually fire.

## Test plan

- [ ] Manually `workflow_dispatch` `sync-rolling-tags` on the branch
      and confirm it still updates rolling tags (and now does not
      touch `.github/workflows/*.yml`).
- [ ] Cut a test release (or wait for the next real release) and
      confirm `sync-rolling-tags` fires automatically on
      `release: published`.
- [ ] Delete a release and confirm `sync-rolling-tags` fires on
      `release: deleted` and re-points / removes the affected rolling
      tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -446,6 +446,11 @@ jobs:
       - name: Create GitHub release
         uses: softprops/action-gh-release@b4309332981a82ec1c5618f44dd2e27cc8bfbfda # v3.0.0
         with:
+          # Use the GitHub App token so that the resulting
+          # `release: published` event triggers downstream workflows
+          # (e.g. sync-rolling-tags.yml). Releases created with the
+          # default GITHUB_TOKEN do not fire downstream workflow events.
+          token: ${{ steps.app-token.outputs.token }}
           tag_name: v${{ needs.prepare-release.outputs.release-version }}
           name: ${{ needs.prepare-release.outputs.release-version }}
           body_path: ${{ needs.prepare-release.outputs.release-notes-filename }}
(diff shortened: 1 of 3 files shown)


s0undt3ch:

- #279 ◀

Issue #281, opened 2026-05-28
`setup-toolr` action fails with `release not found` when pinned by SHA without `version:` input; release prep should also bake the version into `action.yml`


s0undt3ch:
## Symptom

When a caller workflow pins the `setup-toolr` action by commit SHA (as Renovate / Dependabot do by default) and does not pass a `version:` input, the action fails:

```yaml
- uses: s0undt3ch/ToolR@3b6f852b8acd6efc08054ba8d4211cd90c7bdd36
  with:
    cache-prefix: setup-toolr
    cache-tools-venv: true
```

```
setup-toolr: no version input; using action ref '3b6f852b8acd6efc08054ba8d4211cd90c7bdd36'
...
VERSION: 3b6f852b8acd6efc08054ba8d4211cd90c7bdd36
ARCHIVE: toolr-3b6f852b8acd6efc08054ba8d4211cd90c7bdd36-x86_64-unknown-linux-gnu.tar.gz
release not found
Error: Process completed with exit code 1.
```

Real-world hit in the dashtastic repo (PaddleHQ), pinned to the `v0.20.1` release commit via Renovate.

## Root cause

`action.yml`'s "Resolve toolr version" step (lines 85-93) treats `github.action_ref` as a release tag when no `version:` input is provided:

```bash
raw="${INPUT_VERSION}"
if [ -z "$raw" ] && [ -n "${ACTION_REF}" ]; then
  raw="${ACTION_REF}"
  echo "setup-toolr: no version input; using action ref '$raw'"
fi
```

That assumption holds for `uses: …@v0.20.1` but breaks for:

- `@<40-char SHA>` — what Dependabot / Renovate emit by default (with a comment like `# v0.20.1`).
- `@main` / any branch name.
- Custom tags that aren't release tags.

The script then asks `gh release download "v<sha>"`, which doesn't exist → `release not found`.

### 1. Defensive resolve in `action.yml`

When `ACTION_REF` doesn't look like `v?<semver>`, don't blindly use it as a release version. Two acceptable shapes:

- **Easy guard:** if `ACTION_REF` matches `^[0-9a-f]{40}$` (SHA) or doesn't match `^v?[0-9]+\.[0-9]+\.[0-9]+`, fall through to `latest` instead of using it as the version.
[...]

PR #282, opened 2026-05-28
fix(uv): detect musl libc in auto-installer and add actionable exec-failure hint

### LINKED ISSUE #280: Auto-installed uv fails to exec on musl hosts (`I/O error: No such file or directory` from `probe()`)
## Symptom

On a host without `uv` on PATH, with `TOOLR_AUTO_INSTALL_UV=1`, the very first toolr command that needs a venv fails with:

```
toolr: I/O error: No such file or directory (os error 2)
```

Real-world reproducer (Jenkins, musl-libc agent, toolr 0.20.1 installed via `pip install toolr`):

```
+ python3 -m venv .../.toolr-venv
+ .../.toolr-venv/bin/pip install --quiet 'toolr==0.20.1'
+ toolr --version
toolr 0.20.1
+ toolr project deps sync
toolr: I/O error: No such file or directory (os error 2)
```

## Root cause

Two compounding issues in `crates/toolr-core/src/uv/`.

### 1. `host_asset()` always picks the glibc Linux asset

`crates/toolr-core/src/uv/install.rs:105-113` maps `(OS, ARCH)` straight to a release asset:

```rust
("linux", "x86_64") => Some(("uv-x86_64-unknown-linux-gnu", "tar.gz")),
("linux", "aarch64") => Some(("uv-aarch64-unknown-linux-gnu", "tar.gz")),
```

No libc detection. On Alpine / distroless-musl / any musl-libc Linux host, the GNU/glibc `uv` binary is downloaded, written to `$XDG_DATA_HOME/toolr/bin/uv`, `chmod +x`'d, and then handed to `probe()`. The binary's `PT_INTERP` references `/lib64/ld-linux-x86-64.so.2`, which doesn't exist on musl hosts — `execve` returns `ENOENT`.

The `setup-toolr` GitHub Action already does this detection (`action.yml:139-142`):

```bash
libc=gnu
[ -f /etc/alpine-release ] && libc=musl
```

The embedded Rust auto-installer needs the same branch.

### 2. `probe()` returns a context-free `UvError::Io`

`crates/toolr-core/src/uv/discover.rs:32`:

```rust
let output = Command::new(path).arg("--version").output()?;
```

The `?` propagates a bare `io::Error` into `UvError::Io(#[from] io::Error)`, whose `Display` is `"I/O error: {0}"`. `deps_sync()` then collapses the whole chain via `map_err(|e| anyhow::anyhow!(e.user_message()))`, so the user sees just `toolr: I/O error: No such file or directory (os error 2)` — no path, no hint, no recovery action.

`dispatch.rs:195-207` already does the same kind of pre-check for the Python interpreter and emits a clear, actionable error. The uv probe needs equivalent treatment.

## Proposed fix

1. **Detect musl in `host_asset()`** — check `/etc/alpine-release` (cheap, matches what the GH Action does); optionally also parse `ldd --version` stderr for "musl" as a fallback for non-Alpine musl distros. Add the four missing assets: `uv-{x86_64,aarch64}-unknown-linux-musl`.

2. **Wrap `probe()` failures with context.** When `Command::output()` fails with `ErrorKind::NotFound`, return a richer error that:
   - names the candidate path (`/home/foo/.local/share/toolr/bin/uv`),
   - explains that the binary exists but couldn't be executed,
   - hints at the most likely cause (libc mismatch / missing dynamic loader) and a recovery action (install uv via the OS package manager / `pip install uv`).

3. **Don't flatten the error chain in `deps_sync()`.** Replace `map_err(|e| anyhow::anyhow!(e.user_message()))` with a regular `?` + `.context()` so the caller sees the layered error.

## Workaround for users today

- Install `uv` manually on the host (`pip install uv` into the same venv, or via the OS package manager), and **do not** set `TOOLR_AUTO_INSTALL_UV=1`. That bypasses the auto-installer entirely and `find_uv_on_path()` finds the working binary directly.

## Related code

- `crates/toolr-core/src/uv/install.rs:105-136`
- `crates/toolr-core/src/uv/discover.rs:31-52`
- `crates/toolr/src/project.rs:210-222`
- `crates/toolr/src/dispatch.rs:195-207` (pattern to mirror)
- `action.yml:139-150` (libc detection reference)

s0undt3ch:
Closes #280. Related action.yml follow-ups tracked separately: #281 (SHA-pin handling), #283 (auto-install uv via `astral-sh/setup-uv`).

## Summary

When toolr's auto-installer downloads uv on a musl host (Alpine,
distroless-musl, etc.), `host_asset()` previously hard-coded the
glibc Linux asset. The downloaded binary expects
`/lib64/ld-linux-x86-64.so.2`, which doesn't exist on musl, so
`execve` returns `ENOENT`. The error then bottoms out as the
unhelpful `toolr: I/O error: No such file or directory (os error 2)`
because (a) the variant is the generic `UvError::Io` with no
context, and (b) `ensure_venv_ready` flattens the chain through
`anyhow::anyhow!(e.user_message())`. Real-world reproducer: Jenkins
on a musl agent with `TOOLR_AUTO_INSTALL_UV=1`.

This PR is scoped to toolr's *own* Rust auto-installer
(`crates/toolr-core/src/uv/`). The `setup-toolr` GitHub Action
drives `uv sync` directly and never goes through this code path —
its uv-acquisition story is the subject of #283.

## What changed

- **Detect musl in `host_asset()`** (`crates/toolr-core/src/uv/install.rs`)
  New `LinuxLibc` enum + `detect_linux_libc()` with precedence:
  1. `TOOLR_UV_LIBC=musl|gnu` env override — escape hatch for hosts
     where automatic detection guesses wrong.
  2. `/etc/alpine-release` exists → musl. Cheap, matches what
     `setup-toolr` action.yml already does.
  3. `ldd --version` mentions `musl` → musl. Catches non-Alpine
     musl distros.
  4. Default → gnu.
  Asset table extended with the four missing musl entries
  (`uv-{x86_64,aarch64}-unknown-linux-musl`). Detection is isolated
  behind `detect_linux_libc_from()` so unit tests exercise every row
  deterministically.

- **Add `UvError::ExecFailed { path, source }`** (`crates/toolr-core/src/uv/mod.rs`)
  `probe()` now tags `Command::output()` failures with the candidate
  path, distinct from generic `Io`. `user_message()` renders a
  libc-mismatch hint for the ENOENT case naming both the env
  override (`TOOLR_UV_LIBC`) and the OS-package fallback so the user
  has at least two paths forward.

- **Stop flattening the error chain** (`crates/toolr-core/src/project.rs`,
  `crates/toolr/src/dispatch.rs`)
  New `UvError::into_anyhow()` helper preserves the chain, attaching
  `user_message` as a context layer only when it adds information
  beyond `Display`. Replaces the two existing
  `anyhow::anyhow!(e.user_message())` call sites.

## User-visible result

On a musl host hitting the ENOENT path the user now sees:

```
toolr: uv exists on disk but couldn't be executed. This usually
means the binary's dynamic loader is missing — for example a
glibc-linked uv on a musl host, or vice versa. Set
TOOLR_UV_LIBC=musl (or =gnu) to override toolr's libc detection,
or install uv via your OS package manager.: failed to execute uv at
/home/.../toolr/bin/uv: <io::Error>
```

…instead of `toolr: I/O error: No such file or directory (os error 2)`.

## Out of scope

- `setup-toolr` action.yml's missing-uv handling — tracked in #283.
- `setup-toolr` action.yml's SHA-pin resolve bug — tracked in #281.
- Both will be addressed in separate PRs.

## Test plan

- [x] `cargo test --workspace` — green (450+ toolr-core, 7+2 toolr).
- [x] `cargo clippy -p toolr-core --lib -- -D warnings` — clean.
- [x] `cargo clippy -p toolr -- -D warnings` — clean.
- [ ] Manual verify on a musl host (or Docker `alpine:latest`) that
      `TOOLR_AUTO_INSTALL_UV=1 toolr project deps sync` succeeds
      (musl asset downloaded + execs), or surfaces the new
      libc-mismatch hint if detection misses for any reason.
- [ ] Manual verify the `TOOLR_UV_LIBC=musl` / `=gnu` env override
      forces the chosen libc regardless of host detection.

### DIFF
--- a/crates/toolr-core/src/project.rs
+++ b/crates/toolr-core/src/project.rs
@@ -5,7 +5,7 @@ use std::path::Path;
 use anyhow::{Context, Result};
 
 use crate::discovery::discover_project_root;
-use crate::uv::{UvBinary, ensure_uv, install::ConsentMode};
+use crate::uv::{UvBinary, UvError, ensure_uv, install::ConsentMode};
 use crate::venv::{
     ResolvedVenv, perform_editable_installs, resolve_venv_path,
     sync::sync_if_needed, validate::validate_venv, warn_failures,
@@ -22,7 +22,7 @@ pub fn ensure_venv_ready(
         .context("locating project root for the tools venv")?;
     let resolved = resolve_venv_path(&repo_root)
         .context("resolving the tools venv path")?;
-    let uv = ensure_uv(consent).map_err(|e| anyhow::anyhow!(e.user_message()))?;
+    let uv = ensure_uv(consent).map_err(UvError::into_anyhow)?;
     let tools = repo_root.join("tools");
     sync_if_needed(&uv, &tools, &resolved, force_sync)
         .with_context(|| format!("uv sync against {}", tools.display()))?;
--- a/crates/toolr-core/src/uv/discover.rs
+++ b/crates/toolr-core/src/uv/discover.rs
@@ -29,7 +29,15 @@ pub fn find_managed_uv() -> Result<Option<UvBinary>, UvError> {
 
 /// Run `<path> --version`, parse output, validate against the minimum.
 pub fn probe(path: &Path, source: UvSource) -> Result<UvBinary, UvError> {
-    let output = Command::new(path).arg("--version").output()?;
+    // `Command::output()` returns `ErrorKind::NotFound` when the binary
+    // exists on disk but `execve` reports ENOENT — almost always a libc
+    // mismatch (missing dynamic loader). Tag that distinctly so the
+    // user-facing error spells out the recovery path instead of
+    // bottoming out as "I/O error: No such file or directory".
+    let output = Command::new(path)
+        .arg("--version")
+        .output()
+        .map_err(|e| UvError::exec_failed(path, e))?;
     if !output.status.success() {
         return Err(UvError::UnparsableVersion(
             String::from_utf8_lossy(&output.stderr).into_owned(),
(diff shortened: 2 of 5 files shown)


s0undt3ch:

- #282 ◀

PR #284, opened 2026-05-28
fix(action): bake released version into action.yml so SHA pins resolve correctly

### LINKED ISSUE #281: `setup-toolr` action fails with `release not found` when pinned by SHA without `version:` input; release prep should also bake the version into `action.yml`

s0undt3ch:
Closes #281.

## Summary

When a caller workflow pinned the action by SHA without passing a `version:` input, the resolve step treated `github.action_ref` (= the SHA) as the release version and asked `gh release download v<sha>` → opaque `release not found`. Renovate and Dependabot both pin by SHA by default, so callers using either hit this on every dependency bump.

## Fix

Leverage the composite action's own `inputs.version.default`. When that default is kept in sync with the released version, callers who pin by SHA without `version:` input receive the matching binary version automatically — no SHA → tag API lookup needed at action-run time, no extra `gh api` calls, no `GH_TOKEN` dependency for the resolve step.

## What changed

- **`tools/version.py`** — new `_set_action_yml_default_version` helper. Surgically rewrites `inputs.version.default` in action.yml via a single targeted regex (anchored on the 2-space-indent `version:` key + its next `default:` line). Wired into the `bump` command behind a new `--include-action` flag — the bake-in is **opt-in** and only fires in release prep. `bump --write` alone (dev builds, CI branches) leaves action.yml untouched. Dev versions (containing `-`, e.g. `0.21.1-dev42`) are also skipped at the helper level as a safety net. Hard-fails when the regex can't locate the block, so a future structural change to action.yml halts the release workflow loudly instead of silently regressing the bake-in.
- **`.github/workflows/_prepare-release.yml`** — the release-prep `bump` invocation now passes `--include-action`, so action.yml gets bumped alongside Cargo.toml on every release.
- **`action.yml`** — bootstrap `inputs.version.default` to `0.20.1` so the mechanism takes effect immediately for anyone pinning at this commit's SHA. Resolve step simplified back to its pre-#281 shape minus the dead `ACTION_REF` fallback (`inputs.version` is never empty post-bake-in).
- **Tests** — 13 new tests across `tests/test_tools_version.py` (four for the bake-in helper) and `tests/action_yml/test_resolve_version.py` (resolve scenarios + a guard asserting `inputs.version.default` is non-empty so a regression of the bake-in fails CI).

## Caveat

This fixes the issue going forward. Pre-existing pins to old release SHAs (where action.yml-at-that-SHA still has the empty default) still need a `version:` input as a workaround. Anyone bumping their pin to this or a later SHA stops needing the workaround.

## Test plan

- [x] `pytest tests/test_tools_version.py tests/action_yml/` — 19 passed.
- [x] `prek run` — clean (ruff format, ruff check, mypy, actionlint, typos, codespell, hardcoded-secrets).
- [ ] Manual verify: run `toolr version bump --write --include-action --new-version 0.21.0` locally and confirm action.yml's `inputs.version.default` becomes `"0.21.0"` alongside the Cargo.toml bump.
- [ ] Manual verify: run `toolr version bump --write --new-version 0.21.0` (without `--include-action`) and confirm action.yml is **not** touched — the opt-in gate works.
- [ ] Confirm on the next real release that the post-release action.yml in `main` has the correct default.


s0undt3ch:

- #284 ◀

PR #285, opened 2026-05-28
feat(action): install uv via astral-sh/setup-uv when missing or version-pinned

### LINKED ISSUE #283: `setup-toolr` action should install uv via `astral-sh/setup-uv` when missing (or when caller pins a version)
## Symptom

The `setup-toolr` action requires `uv` on PATH but doesn't install it. When a caller workflow doesn't set up uv first (no `mise install uv`, no `astral-sh/setup-uv`, etc.), the "Sync tools venv" step bails:

```
setup-toolr: 'uv' is required on PATH to materialise the tools venv
setup-toolr: install via mise or your platform's uv channel
Error: Process completed with exit code 1.
```

Real-world hit in dashtastic (PaddleHQ) — caller workflow installs everything via mise except uv. The action just got far enough to download the toolr binary and resolve the in-tree venv path before tripping on the missing uv.

## Source

`action.yml:389-393`:

```bash
if ! command -v uv >/dev/null 2>&1; then
  echo "setup-toolr: 'uv' is required on PATH to materialise the tools venv" >&2
  echo "setup-toolr: install via mise or your platform's uv channel" >&2
  exit 1
fi
```

The action drives `uv sync` directly (for good reasons — `--no-default-groups --frozen` scopes the install to what CI actually needs), so it can't fall back to `toolr project deps sync` which would invoke toolr's own auto-installer.

## Proposed design

1. **New input** `uv-version` on `setup-toolr`. Default `""`. Passed through to `astral-sh/setup-uv`'s `version` input when invoked. Lets callers pin a specific uv version when they need to; empty means "whatever setup-uv installs by default".

2. **New step** in `action.yml`, between "Install toolr binary" and "Sync tools venv":

   ```yaml
   - name: Detect uv on PATH
     id: uv-check
     shell: bash
     run: |
       if command -v uv >/dev/null 2>&1; then
         echo "found=true" >> "$GITHUB_OUTPUT"
       else
         echo "found=false" >> "$GITHUB_OUTPUT"
       fi

   - name: Install uv via astral-sh/setup-uv
     if: steps.uv-check.outputs.found == 'false' || inputs.uv-version != ''
     uses: astral-sh/setup-uv@<pinned-sha>
     with:
       version: ${{ inputs.uv-version }}
   ```

   Trigger logic (matches the agreed design):
   - **uv missing AND no `uv-version`** → run setup-uv with no version pin (installs setup-uv's default).
   - **uv missing AND `uv-version` set** → run setup-uv pinned to that version.
   - **uv on PATH AND `uv-version` set** → run setup-uv anyway, overriding whatever's on PATH. Lets callers force a specific version even when mise / a prior step put a different one on PATH.
   - **uv on PATH AND no `uv-version`** → skip setup-uv. Respect what the caller already provided.

3. **Sync-tools-venv step** keeps its `command -v uv` guard as a sanity assertion (should never fail post-fix), or drops it entirely. Either is fine; keeping it adds one cheap line and a clearer error if the new logic ever regresses.

4. **Use `astral-sh/setup-uv` rather than rolling our own** because:
   - Official, well-maintained, handles caching transparently.
   - Supports version pinning, attestation verification.
   - Reusing it costs us one pinned SHA in the workflow and gives us all of that for free.

   Alternative considered: `toolr self install-uv` driven by the action. Rejected because (a) we'd then need a third actions/cache step keyed on toolr version to avoid re-downloading uv every run, and (b) we'd duplicate caching logic that setup-uv already gets right.

## Workaround for callers today

Install uv before invoking `setup-toolr`. Either:

```yaml
- uses: astral-sh/setup-uv@<sha>
- uses: s0undt3ch/ToolR@<sha>
  with:
    version: 0.20.1
```

…or install via mise / pipx / your platform's uv channel earlier in the workflow.

## Related

- #281 (separate `action.yml` bug — SHA-pin handling without `version:` input)
- #280 (separate but related — toolr's *own* auto-installer needs musl detection + better error messages; doesn't apply here because the action drives uv directly, not through toolr)

s0undt3ch:
Closes #283.

## Summary

The `setup-toolr` action required `uv` on PATH but never installed it. Caller workflows that didn't pre-install uv (via mise, an upstream `astral-sh/setup-uv` step, or an OS package) hit a hard error at the "Sync tools venv" step with no recovery short of editing the caller's workflow. This adds a `uv-version` input and an internal `astral-sh/setup-uv` step so the action is zero-prerequisite by default while still letting callers pin or override uv.

## Truth table

| uv on PATH? | `uv-version` input | Behavior |
|-------------|--------------------|----------|
| No          | (empty)            | setup-uv installs its own default version. |
| No          | `0.5.1`            | setup-uv installs uv 0.5.1. |
| Yes         | (empty)            | Skip setup-uv — respect caller's uv. |
| Yes         | `0.5.1`            | Run setup-uv anyway — caller explicitly asked for a specific version. |

## What changed

- **New `uv-version` input** on the composite action, default `""`. Passed through verbatim to `astral-sh/setup-uv`'s `version` input.
- **New "Detect uv on PATH" step** records whether uv is already available (`steps.uv-check.outputs.found`).
- **New "Install uv via astral-sh/setup-uv" step** gated on `steps.uv-check.outputs.found == 'false' || inputs.uv-version != ''`. Pinned to `astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b` (v8.1.0).
- **Sync tools venv step** keeps its `command -v uv` guard, but the error message is reworded — reaching it now means the setup-uv step skipped and uv has since disappeared. Almost certainly a bug, not a missing prerequisite.

## Why `astral-sh/setup-uv` rather than rolling our own

Caching, attestation verification, and version pinning are all handled by the upstream action — that's the whole reason for choosing it over driving `toolr self install-uv` from the action (which would force us to reinvent all three).

## Test plan

- [x] `prek run --files action.yml` — clean.
- [ ] Manual verify on a runner without uv pre-installed: `setup-toolr` should now install uv and succeed all the way through "Sync tools venv" without any caller-side prep.
- [ ] Manual verify on a runner with uv already on PATH and no `uv-version` input: setup-uv step should be skipped (visible in the action log) and the caller's uv used.
- [ ] Manual verify `uv-version: latest` and `uv-version: 0.5.1` both force a setup-uv install even when uv is already on PATH.

## Related

- #281 / #284 — separate `action.yml` work for SHA-pin handling (bake-in mechanism).
- #280 / #282 — toolr's own Rust auto-installer (musl detection + better errors). Unrelated to this PR because the action drives uv directly, not via toolr.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -63,6 +63,18 @@ inputs:
       would skip (e.g. `--group lint`). Defaults to empty.
     default: ""
 
+  uv-version:
+    description: |
+      uv version to install via `astral-sh/setup-uv` when uv isn't
+      already on PATH (or when this input is non-empty, in which
+      case setup-uv runs regardless and overrides whatever's on
+      PATH). Accepts anything `astral-sh/setup-uv` accepts —
+      bare semver (`0.5.1`), with leading `v` (`v0.5.1`), or
+      `latest`. Empty (the default) means "no version pin": if uv
+      is missing setup-uv installs its own default; if uv is
+      already on PATH the action leaves it alone.
+    default: ""
+
 outputs:
   version:
     description: Resolved ToolR version installed (no leading `v`).
@@ -317,6 +329,35 @@ runs:
         } >> "$GITHUB_OUTPUT"
         echo "setup-toolr: installed toolr ${VERSION} at ${bin_dir}/${BIN}"
 
+    # uv is required for the "Sync tools venv" step further below.
+    # Detect it first so we only invoke `astral-sh/setup-uv` when
+    # necessary — when the caller's environment already provides uv
+    # (mise, a prior setup-uv step, an OS package), respect that.
+    # The exception is `inputs.uv-version`: a non-empty value means
+    # "give me exactly this uv version", so we run setup-uv to
+    # install it regardless of what's already on PATH.
+    - name: Detect uv on PATH
+      id: uv-check
+      shell: bash
+      run: |
+        if command -v uv >/dev/null 2>&1; then
+          echo "found=true" >> "$GITHUB_OUTPUT"
+          echo "setup-toolr: found uv on PATH at $(command -v uv)"
+        else
+          echo "found=false" >> "$GITHUB_OUTPUT"
+          echo "setup-toolr: uv not found on PATH"
+        fi
+
+    - name: Install uv via astral-sh/setup-uv
+      if: steps.uv-check.outputs.found == 'false' || inputs.uv-version != ''
+      uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+      with:
+        # Empty `version` lets setup-uv pick its own default — that's
+        # exactly the behaviour we want for "uv is just missing, install
+        # something sensible". Callers wanting a specific version pass
+        # it via setup-toolr's `uv-version` input.
+        version: ${{ inputs.uv-version }}
+
     - name: Resolve tools venv path
       id: venv-path
       if: inputs.cache-tools-venv == 'true'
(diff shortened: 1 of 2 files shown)


s0undt3ch:

- #285 ◀

PR #293, opened 2026-06-02
fix(parser): split command_group docstring into title + description

### LINKED ISSUE #292: command_group(docstring=...) does not split first line into title; group description never appears in parent --help
## Summary

`command_group(\"name\", docstring=__doc__)` does not parse the docstring into a short/long
pair the way `@command` does. The whole docstring lands in the group's `description`
field, and `title` stays empty — so the group's blurb never appears next to its name in
the parent `--help` listing, and `toolr <group> -h` (clap's short help) also shows
nothing.

## Reproduce

```python
# tools/dbt_config.py
\"\"\"Paddle's dbt config and setup routines.\"\"\"

from toolr import command_group

group = command_group(\"dbt-config\", docstring=__doc__)
```

Resulting manifest entry:

```json
{
  \"groups\": [
    { \"name\": \"dbt-config\", \"title\": \"\", \"description\": \"Paddle's dbt config and setup routines.\" }
  ]
}
```

Observed:

```text
$ toolr -h
Commands:
  dbt-config                                ← no blurb (title is empty)
  project     Operations on the current repo's tools/ directory
  self        Operations on toolr itself

$ toolr dbt-config -h
Usage: toolr dbt-config [COMMAND]          ← no blurb (about == title == \"\")
...
```

`toolr dbt-config --help` (long form) does render the description because the static
parser stuffed the whole docstring into `description`, which clap then renders as
`long_about`. So the text *is* plumbed through, just into the wrong slot.

## Root cause

Two callsites, both wrong in different ways. They share the same underlying parser
(`crates/toolr-core/src/docstrings.rs` — exposed to Rust as `SimpleDocstringParser`
and to Python via PyO3 as `DocstringParser` / `Docstring.parse`):

1. **Static parser** — `crates/toolr-core/src/parser/groups.rs::parse_group_call`
   never runs the docstring through `SimpleDocstringParser`. `parser/commands.rs`
   does the split correctly (`summary` = `short_description`, `description` =
   `long_description`); `parser/groups.rs` skips the step entirely.

2. **Runtime decorator** — `crates/toolr-py/python/toolr/_decorators.py::command_group`
   *does* call `Docstring.parse(...)` but then assigns:
   - `description = parsed.short_description`  (first line → wrong field)
   - `long_description = parsed.long_description`  (rest → field that's never persisted)

   It also defaults `title = name` when no title was given, which duplicates the
   leaf name redundantly in clap output (the static parser doesn't do this fallback,
   so the two paths disagree).

3. **Manifest schema** — `crates/toolr-core/src/manifest/model.rs::Group` only carries
   `title` + `description`; there's no `long_description` field, so the Python
   runtime's `long_description` would be silently dropped at manifest-emit time even
   if it were populated correctly.

## Expected

Mirror the command-parsing convention from `parser/commands.rs`:

- **`title`** ← first paragraph of the docstring (`short_description`), unless an
  explicit positional title was passed.
- **`description`** ← remainder of the docstring (`long_description`), unless an
  explicit `description=` / 3rd-positional was passed.

Then `toolr -h` would render:

```text
Commands:
  dbt-config  Paddle's dbt config and setup routines.
  ...
```

And `toolr dbt-config -h` / `--help` would each show the right slice.

## Suggested fix

1. `parser/groups.rs::parse_group_call` — run the `docstring=` kwarg through
   `SimpleDocstringParser`; populate `title` from `short_description` (when no
   positional title was provided) and `description` from `long_description`.
2. `_decorators.py::command_group` — when `docstring=` is set: assign
   `short_description` → `title` (if not explicit), `long_description` → `description`.
   Drop the `if not title: title = name` fallback.
3. Update the public `command_group` docstring + `references/commands.md` so the
   contract reads \"first line → title, rest → description\".

s0undt3ch:
## Summary

Fixes [#292](https://github.com/s0undt3ch/ToolR/issues/292) — `command_group("name", docstring=__doc__)` now splits the docstring into `title` + `description` the same way `@command` splits function docstrings.

- **Static parser** (`crates/toolr-core/src/parser/groups.rs`): the `docstring=` kwarg is run through the shared parser. First paragraph → `title` (clap's `about`, shown in the parent's command listing). Remainder → `description` (clap's `long_about`). Explicit positional `title=` / `description=` still win.
- **Runtime decorator** (`toolr._decorators.command_group`): mirrors the same split so runtime introspection matches the static manifest. Drops the `title = name` fallback that would otherwise produce a redundant `dbt-config  dbt-config` row in `toolr -h`.
- **Shared helper** (`parser/mod.rs::parse_docstring`): both `parser/groups.rs` and `parser/commands.rs` now go through this single helper, so the "first paragraph → short, rest → long" convention lives in one place.

The skill-authoring example manifest snapshot was regenerated to reflect the new (correct) titles, and the public docstring + `references/commands.md` were updated.

### Fixture cleanup (second commit)

While in the area, the `del clean_registry` / `del isolated_sys_path` pattern in `tests/decorators/test_decorators_unit.py` and `tests/introspect/test_introspect_unit.py` is gone. The decorators file makes `clean_registry` `autouse=True` (every test in it needs the snapshot/restore isolation); the introspect file uses `@pytest.mark.usefixtures(...)` where only some tests need each fixture. 22 `del <fixture>` lines removed; no behavioural change to the tests.

## Test plan

- [x] `cargo test -p toolr-core` — 470 + 4 + 1 + 0 passing.
- [x] New unit tests in `parser/groups.rs`: short docstring → title, multi-paragraph docstring → title + description, explicit `description=` overrides docstring-long, explicit positional `title` overrides docstring-short.
- [x] New / updated decorator tests in `tests/decorators/test_decorators_unit.py` and `tests/registry/test_command_group_errors.py` cover the same matrix from the Python side.
- [x] Skill-examples snapshot regenerated (`cargo test -p toolr-core --test skill_authoring_examples -- --ignored regenerate`) and re-asserted.
- [x] `pytest tests/decorators tests/registry tests/introspect tests/runner tests/sources` — 165 passing.
- [x] `prek run --files ...` — full local lint/typecheck/format chain passes.
- [ ] CI green.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,3 +10,26 @@ this file to empty for the next cycle.
 Empty between releases is the steady-state — there's no header,
 no scaffolding. Just write whatever should appear in the notes.
 -->
+
+## Bug fixes
+
+### `command_group(docstring=...)` now splits the docstring into title + description
+
+`command_group("name", docstring=__doc__)` previously stuffed the entire
+docstring into the group's `description` field and left `title` empty,
+so the blurb never appeared next to the group in the parent `--help`
+listing (clap shows `about`, which is what `title` populates). The
+parser now mirrors how `@command` function docstrings are handled — the
+first paragraph becomes the group's `title` (clap's `about`) and the
+remainder becomes the `description` (clap's `long_about`). An explicit
+positional `title=` / `description=` still wins for either side.
+
+Both the static parser (`parser/groups.rs`) and the runtime decorator
+(`toolr._decorators.command_group`) now go through the same shared
+`parse_docstring` helper, so static-manifest output and runtime
+introspection agree. The runtime's old `title = name` fallback —
+which would have produced a redundant `dbt-config  dbt-config` in
+the parent listing — has been removed; an unset title now stays
+empty, matching the static parser.
+
+See [#292](https://github.com/s0undt3ch/ToolR/issues/292).
--- a/crates/toolr-core/src/parser/commands.rs
+++ b/crates/toolr-core/src/parser/commands.rs
@@ -5,12 +5,12 @@ use std::collections::HashMap;
 use ruff_python_ast::{Decorator, Expr, ModModule, Stmt, StmtFunctionDef};
 
 use super::groups::GroupBinding;
+use super::parse_docstring;
 use super::signatures::extract_arguments;
 use super::symbols::ArgSectionTable;
 use super::symbols::EnumTable;
 use super::symbols::TypeAliasTable;
 use super::types::{SourcesImports, TypeImports, TypeResolutionError, resolve_arguments};
-use crate::SimpleDocstringParser;
 use crate::manifest::{Command, Origin};
 
 type GlobalVars = HashMap<String, String>;
@@ -236,7 +236,7 @@ fn build_command(
     errors: &mut Vec<TypeResolutionError>,
 ) -> Command {
     let raw_doc = function_docstring(func);
-    let parsed = SimpleDocstringParser::new().parse(&raw_doc).ok();
+    let parsed = parse_docstring(&raw_doc);
     let summary = parsed
         .as_ref()
         .map(|d| d.short_description.clone())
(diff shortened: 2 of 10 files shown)


s0undt3ch:

- #293 ◀

PR #305, opened 2026-06-02
fix(coverage): rewrite venv paths back to source; ratchet codecov per-crate

s0undt3ch:
## Summary

Two coupled coverage infrastructure fixes:

### 1. `.coveragerc`: rewrite venv-installed paths back to source

Found while investigating why `toolr-py` shows 25.8% project coverage on main when local runs of the same test suite hit 90%+ on every affected file:

CI installs `toolr-py` from a prebuilt wheel into `.venv/lib/python*/site-packages/toolr/`. The existing `[run] source = crates/toolr-py/python/` filter dropped every hit landing on the wheel-installed copy as "not in source", so the CI `coverage report` literally showed:

```
crates/toolr-py/python/toolr/_decorators.py     99    99   28   0    0%
crates/toolr-py/python/toolr/_introspect.py     66    66   14   0    0%
crates/toolr-py/python/toolr/_runner.py        229   229   58   0    0%
```

— despite the tests for those modules **passing**. Local `maturin develop` uses an editable install pointing back at the source tree, so the bug was invisible locally.

A `[paths]` aliases block in `.coveragerc` maps the venv install location back onto the canonical source path during `coverage combine`. Local coverage drops in *no* hits (the editable install already produces the canonical path); CI now sees the actual coverage.

### 2. `.codecov.yml`: tighten the gate so coverage can't quietly slide

- `coverage.status.project.default`: `target: 90%` / `threshold: 1%` → `target: auto` / `threshold: 0%`. Overall coverage can never drop on a PR; the bar ratchets up implicitly. The 1% wobble was tolerating slow decay.
- `coverage.status.patch.default`: `target: 90%` → `target: 99%`. New code on every PR must hit 99% of its own lines.
- New `component_management` section pins per-crate baselines so a regression inside the weakest crate (e.g. `toolr-py`) can't be masked by a gain elsewhere. Components: `toolr-core`, `toolr`, `toolr-py`, `xtask`, each with `target: auto` / `threshold: 0%`.
- Comment layout gains a `components` block.

## Expected outcome

After this PR merges, the next codecov run should show `toolr-py` jumping from 25.8% to something in the 80-90% range — the same coverage local runs always saw, just no longer silently filtered out. The new per-crate auto-ratchet pins whatever that real number turns out to be as the floor.

This is **honest** coverage — the tests were already there and passing; we just weren't counting their hits.

## Test plan

- [x] `curl --data-binary @.codecov.yml https://codecov.io/validate` returns "Valid!".
- [x] `.coveragerc` still parses; local `coverage report` against the existing combined data file produces correct numbers.
- [ ] CI green; codecov shows the expected jump in `toolr-py` coverage.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -16,11 +16,14 @@ coverage:
   status:
     project:
       default:
-        # Block PR merges if total coverage drops below 90%.
-        target: 90%
-        # Allow a 1% wobble for noise (random scheduler differences in
-        # signal-driven tests, etc.) without auto-failing the status.
-        threshold: 1%
+        # `auto` pins the bar to the base commit's coverage — any drop
+        # in overall project coverage on a PR fails the status. The
+        # bar ratchets up implicitly: a PR that raises coverage also
+        # raises the floor for the next PR.
+        target: auto
+        # No wobble. With `auto`, a 1% drop is meaningful and should
+        # be visible — there's no "noise" baseline to absorb.
+        threshold: 0%
         # Wait until every flag has reported before computing the status.
         # We upload 8 flags (4 Python × 2 OSes); without this Codecov
         # sometimes flips green/red as the partial uploads land.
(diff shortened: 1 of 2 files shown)


s0undt3ch:

- #305 ◀

PR #307, opened 2026-06-03
fix(complete): always offer `self` / `project` built-ins outside a toolr project

### LINKED ISSUE #306: Tab completion offers nothing outside a toolr project; built-ins should still complete
## Summary

Running tab completion (`toolr __complete <cwd> …`) from a directory that has no `tools/` ancestor returns **nothing at all** and exits 1. The shell then falls back to default completion.

That's wrong for the built-in subtree: regardless of cwd, the binary itself owns `self …`, `project …` (and accepts `--version` / `--help`). Those should still be offered as completion candidates everywhere — only user-defined commands need a project root.

## Reproduction

```console
$ /path/to/toolr __complete /tmp ""
$ echo $?
1
```

…vs. inside a project where the same call lists `bench ci project self version`.

In a shell, this means:

```console
$ cd /some/dir-with-no-tools
$ toolr <TAB>     # → nothing; falls through to filename completion
```

Users expect at minimum:

```
project  self
```

(plus `--version`, `--help`, etc.)

## Root cause

`crates/toolr/src/dispatch.rs::run_complete` (around line 410):

```rust
let Ok(resolved) = resolve_manifest_at_tab(&cwd) else {
    return Ok(ExitCode::from(1));
};
let mut manifest = resolved.manifest;
let (extra_groups, extra_commands) =
    crate::builtin_completions::built_in_completion_entries();
manifest.groups.extend(extra_groups);
manifest.commands.extend(extra_commands);
```

`resolve_manifest_at_tab` errors when `discover_project_root` can't find a `tools/` ancestor (`crates/toolr-core/src/complete/freshness.rs:28-30`). The `let-else` early-returns before the built-in entries are ever injected.

The current test `silent_failure_when_no_tools_dir_anywhere` in `crates/toolr/tests/complete_smoke.rs:133` actually codifies this buggy behaviour (`assert!(output.stdout.is_empty())`), so it needs to be updated alongside the fix.

## Expected behaviour

When `resolve_manifest_at_tab` errors with no project root, fall back to an empty manifest, then still merge in the built-in completion entries. Built-ins always complete; user-defined commands only complete inside a toolr project.

## Affected versions

`main` and at least back to 0.22.x (built-in completion entries were introduced before the freshness rework).

s0undt3ch:
## Summary

Fixes #306.

Running `toolr <TAB>` from a directory with no `tools/` ancestor used to return nothing and exit 1, so the shell fell back to filename completion. The binary's own `self` / `project` subtree doesn't depend on a project root, so it should always complete — only user-defined groups need a discoverable `tools/`.

## Root cause

`crates/toolr/src/dispatch.rs::run_complete` early-returned with `ExitCode::from(1)` whenever `resolve_manifest_at_tab` failed (no `tools/` ancestor). The synthetic built-in entries from `crate::builtin_completions` were never injected, so even completions that don't need a project root (`self …`, `project …`) silently disappeared.

`resolve_manifest_at_tab` (in `toolr_core::complete::freshness`) needing a project root is correct — it manages the on-disk manifest cache. The bug was at the dispatch layer, where the failure was treated as fatal instead of "no user-defined commands, use built-ins only".

## Reproduction (before)

```console
$ cd /tmp                                     # no tools/ anywhere up the tree
$ toolr __complete /tmp ""
$ echo $?
1
```

In a shell: `toolr <TAB>` → nothing.

## After

```console
$ toolr __complete /tmp ""
project
self
$ toolr __complete /tmp self ""
build-manifest
cache
completion
$ toolr __complete /tmp project ""
init
manifest
venv
```

Inside a project, user-defined commands continue to be offered (manifest discovery path unchanged).

## Changes

- `crates/toolr/src/dispatch.rs` — `run_complete` falls back to an empty manifest when `resolve_manifest_at_tab` errors, then merges the built-in entries as before. New `empty_manifest_for_completion` helper.
- `crates/toolr/tests/complete_smoke.rs` — the existing `silent_failure_when_no_tools_dir_anywhere` test asserted the buggy behaviour (exit 1, empty stdout); renamed to `completes_builtins_when_no_tools_dir_anywhere` and rewritten to assert `self` and `project` are offered.
- `UNRELEASED.md` — release-note entry under "Bug fixes".

## Test plan

- [x] `cargo test -p toolr` — all 18 binaries green (including the renamed completion test).
- [x] Manual `toolr __complete` from `/tmp` returns `project` / `self` and exits 0.
- [x] Manual `toolr __complete` inside this repo still returns the user-defined `bench` / `ci` / `version` groups alongside the built-ins.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -39,3 +39,18 @@ Side-effects:
   so the two surfaces can't drift.
 
 See [#292](https://github.com/s0undt3ch/ToolR/issues/292).
+
+## Bug fixes
+
+### Tab completion offers `self` / `project` even outside a toolr project
+
+Running `toolr <TAB>` from a directory with no `tools/` ancestor used to
+return nothing and exit 1, so the shell fell back to filename
+completion. The binary's own `self` / `project` subtree doesn't depend
+on a project root, so it should always complete — only user-defined
+groups need a discoverable `tools/`. `run_complete` now falls back to
+an empty manifest when project discovery fails, then merges in the
+built-in completion entries so `self`, `project`, and their children
+are offered everywhere.
+
+See [#306](https://github.com/s0undt3ch/ToolR/issues/306).
--- a/crates/toolr/src/dispatch.rs
+++ b/crates/toolr/src/dispatch.rs
@@ -13,7 +13,7 @@ use toolr_core::discovery::discover_project_root;
 use toolr_core::execute::{
     resolve_python, spawn_runner, wait_with_signals, write_spec_to_tempfile,
 };
-use toolr_core::manifest::Manifest;
+use toolr_core::manifest::{Manifest, SCHEMA_VERSION};
 use toolr_core::venv::resolve_venv_path;
 
 use crate::execute_build::{OutputOptions, build_dispatch_spec, build_spec, pack_child_args};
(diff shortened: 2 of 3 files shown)


s0undt3ch:

- #307 ◀

PR #308, opened 2026-06-03
docs(agent): add CLAUDE.md to canonicalize agent behavior in this repo

s0undt3ch:
## Summary

Adds a `CLAUDE.md` at the repo root that gives Claude Code (and other agents) a single canonical source of truth for how to work in this repo. Previously the same information was scattered across `CONTRIBUTING.md`, per-user agent memory, and tribal convention — agents working on the repo for the first time had to re-derive it from scratch.

Also gitignores `.claude.local.md` so per-user agent overrides can sit alongside the shared file without leaking into git.

## Structure

Two tiers, intentionally distinct so the file stays scannable as it grows:

### Facts

Reference material an agent loads once and re-reads on demand.

- **Workspace layout** — the four crates and the Python source path under `crates/toolr-py/python/toolr/`.
- **Commands** — mise-driven setup, the `mise run test` umbrella, per-suite runners, dogfood invocations.
- **Where things live** — release-notes queue, **specs (live + archive) with the toolr-specific naming convention**, completion entry points, parser, freshness module, reference plugin, project-local hooks, shipped skills.

### Working rules

Behavior an agent applies while acting. Split three ways:

- **Decisions (when)** — scaling verification to PR scope, monitoring long `cargo test` runs, the brainstorm-then-plan flow.
- **Actions (how)** — `RUNNER_SCHEMA_VERSION`/`SCHEMA_VERSION` lock-step, `UNRELEASED.md` vs `CHANGELOG.md`, regen doc snippets, regen skill refs after public-surface changes, git-spice for stacked PRs, Conventional Commits, no `--no-verify`, pytest factory fixtures, **archive specs as the last implementation step (same PR as implementation, not follow-up)**.
- **Off-limits (what not)** — no `toolr-django` crate, no cross-repo sharing infra, no 1.0 features yet (0.20.0 ships first).

A one-line header makes the file authoritative: *"This file is canonical for agent behavior in this repo. Rules here override the agent's default skills and memory when they conflict."* That's what makes callouts like "Override the brainstorming-skill default of `docs/superpowers/specs/`" actually carry weight.

## Why this matters

- **Specs land in the right place.** The brainstorming skill's default of `docs/superpowers/specs/` doesn't match this repo's `specs/` convention (documented in `specs/README.md`). New designs were at risk of writing to the wrong path until the agent stumbled onto the README. CLAUDE.md surfaces the override on first read.
- **Schema-version drift gets caught before CI.** The `RUNNER_SCHEMA_VERSION` / `SCHEMA_VERSION` pair only stays in sync if everyone touching the runner spec knows to bump both. CONTRIBUTING.md already documents this for humans; CLAUDE.md guarantees agents see it too.
- **Memory-only knowledge gets shared.** Items like "0.20.0 ships before 1.0.0", "no `toolr-django`", "reference plugin is `examples/plugin-package/`", and "scale tests to PR scope" lived only in per-user auto-memory. New collaborators' agents would have re-suggested already-declined ideas. Promoting them into CLAUDE.md fixes that.

## What's deliberately NOT in the file

- The "no `Co-Authored-By: Claude …` trailer" rule is a personal/local convention; it stays in per-user agent memory, not the shared file.
- AGENTS.md / GEMINI.md alternates — single Claude-only file for now; revisit if other agents land.
- Per-crate CLAUDE.md files — the workspace is small enough that one root file covers it.

## Test plan

- [x] `rumdl` passes (no lines > 120, indentation matches MD007).
- [x] `prek run --all-files` passes via pre-commit.
- [x] File renders correctly on GitHub.


s0undt3ch:

- #308 ◀

PR #311, opened 2026-06-04
bench: align markdown separators + run startup bench against built artifacts in CI

s0undt3ch:
## Summary

Two related changes around `toolr bench compare`:

1. **`fix(tools/bench)`** — `_render_markdown` was emitting
   `--- + :` for right-aligned numeric columns (`width` dashes
   plus an alignment colon = `width + 1` visible chars) while
   data cells padded to `width` chars. The separator row was one
   column wider than every data row per numeric column, breaking
   the hand-readable alignment of the raw markdown source.
   Emit `(width - 1)` dashes plus `:` so the separator and data
   rows have the same visible width.

2. **`ci`** — adds three `bench-{linux,macos,windows}` callers
   in `ci.yml` plus a new reusable workflow
   `.github/workflows/_bench.yml` (mirrors the test-* /
   build-* callable-workflow split). Each caller runs
   `toolr bench compare --markdown --install` against THIS PR's
   built `toolr` binary archive + matching `toolr-py` wheel and
   posts the markdown table to its runner's step summary.

### Before (alignment fix)

```text
| First (ms) | Second (ms) |
| ----------: | -----------: |
|       22.3 |        17.9 |
```

### After

```text
| First (ms) | Second (ms) |
| ---------: | ----------: |
|       22.3 |        17.9 |
```

## How the CI bench job gates the build

`bench-{linux,macos,windows}` are listed in
`set-pipeline-exit-status.needs`. A *job failure* (archive
missing, install crash, `toolr` crash, etc.) fails the build.
The benchmark *numbers themselves* never fail the build —
`toolr bench compare` exits zero regardless of how fast each
tool ran, so slow startup numbers won't gate PRs; only broken
benches do.

## Job choreography (`_bench.yml`)

- Single `case "$RUNNER_OS"` step emits `toolr-triple` +
  `wheel-platform` outputs — same pattern as `_test.yml`.
- Archive extraction handles both `.tar.gz` and `.zip`,
  flattening the inner directory on Windows. The standalone
  `toolr` binary lands on PATH for downstream steps.
- Bootstrap the tools venv via `toolr project venv sync`,
  then swap the PyPI-resolved `toolr-py` for the just-built
  wheel via `uv pip install --python <venv> --force-reinstall`.
  The bench measures THIS PR's binary + wheel, not the
  released versions.
- Prepend `uv tool dir --bin` to `$GITHUB_PATH` so the
  `--install`-driven `uv tool install <pkg>` placements for
  invoke/duty/doit/nox/python-tools-scripts are discoverable
  by `_resolve_binary`.
- The cp-tag is derived from the venv's active Python after
  `setup-virtualenv` (mirroring the docs job) so a `mise.toml`
  python bump auto-selects the matching wheel artifact.

## Drive-by

Routes `github.event_name` / `github.ref_name` through `env:`
in `prepare-ci`'s "Check if the build should run" step. Branch
names are attacker-controllable, so interpolating them into a
shell `run:` is a CWE-78-shaped risk; the `toolr` CLI receives
them as ordinary argv either way. Caught by the local Semgrep
hook while authoring this PR.

## Test plan

- [x] `cargo run -p toolr -- bench compare --path "$(mktemp -d)" --markdown --runs 10`
      — separator row is byte-aligned with data rows in the
      output.
- [x] `prek run --files .github/workflows/ci.yml .github/workflows/_bench.yml`
      — actionlint / pin-shas / shellcheck / typos all green.
- [ ] CI: confirm the `Bench (Linux|macOS|Windows)` jobs run,
      produce a populated step summary, and the markdown table
      alignment is correct on every runner.
- [ ] CI: confirm a forced bench failure (e.g. wrong artifact
      name) fails the overall pipeline via
      `set-pipeline-exit-status`.

_claude --resume 9710922d-2b5f-400b-99f5-7db6d6ea8af8_

### DIFF
--- a/.github/workflows/_bench.yml
+++ b/.github/workflows/_bench.yml
@@ -1,0 +1,192 @@
+name: Bench
+
+on:
+  workflow_call:
+    inputs:
+      os:
+        required: true
+        type: string
+        description: 'Operating system to run bench on'
+      display-name:
+        required: true
+        type: string
+        description: 'Display name for the bench job (Linux / macOS / Windows)'
+      cache-seed:
+        required: true
+        type: string
+        description: 'Seed mixed into every cache key. Bump in the caller workflow to invalidate.'
+      mise-version:
+        required: true
+        type: string
+        description: 'Version of mise to install (forwarded to the setup-virtualenv composite). Pass the caller workflow''s MISE_VERSION env var.'
+
+permissions:
+  contents: read
+
+jobs:
+  bench:
+    name: ${{ inputs.display-name }}
+    runs-on: ${{ inputs.os }}
+    permissions:
+      contents: read
+
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      # Derive the artifact identifiers we'll download below from the
+      # runner OS. Centralised here so the rest of the job can reference
+      # `steps.artifacts.outputs.*` instead of repeating the same `case
+      # $RUNNER_OS` everywhere. Mirrors `_test.yml`'s pattern.
+      - name: Compute artifact identifiers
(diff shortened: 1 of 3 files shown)


s0undt3ch:

- #311 ◀

PR #312, opened 2026-06-04
ci: bench `toolr` startup against built artifacts on Linux/macOS/Windows

s0undt3ch:
## Summary

Adds three `bench-{linux,macos,windows}` callers in `ci.yml` plus
a new reusable workflow `.github/workflows/_bench.yml` (mirrors
the test-* / build-* callable-workflow split). Each caller runs
`toolr bench compare --markdown --install` against this PR's
built `toolr` binary archive and matching `toolr-py` wheel and
posts the markdown table to its runner's step summary.

### How it gates the build

`bench-{linux,macos,windows}` are listed in
`set-pipeline-exit-status.needs` — a *job failure* (archive
missing, install crash, `toolr` crash, etc.) fails the build.
The benchmark *numbers themselves* never fail the build: `toolr
bench compare` exits zero regardless of how fast each tool ran,
so slow startup numbers won't gate PRs — only broken benches do.

### Job choreography

- Bootstrap the tools venv via `toolr project venv sync`, then
  swap the PyPI-resolved `toolr-py` for the just-built wheel via
  `uv pip install --python <venv> --force-reinstall`. The bench
  measures THIS PR's binary + wheel, not the released versions.
- Prepend `uv tool dir --bin` to `$GITHUB_PATH` so the
  `--install`-driven `uv tool install <pkg>` placements for
  invoke/duty/doit/nox/python-tools-scripts are discoverable by
  `_resolve_binary`.
- The cp-tag is derived from the venv's active Python after
  `setup-virtualenv` (same as the docs job) so a `mise.toml`
  python bump auto-selects the matching wheel artifact.

`_bench.yml` mirrors `_test.yml` exactly: `workflow_call` with
`os` / `display-name` / `cache-seed` / `mise-version` inputs,
a single `case "$RUNNER_OS"` step emitting `toolr-triple` +
`wheel-platform` outputs, and the same archive-extraction
pattern (handles both `.tar.gz` and `.zip`, flattens the inner
dir on Windows). Subsequent steps reference `toolr` directly
via PATH.

### Drive-by

Routes `github.event_name` / `github.ref_name` through `env:`
in `prepare-ci`'s "Check if the build should run" step. Branch
names are attacker-controllable, so interpolating them into a
shell `run:` is a CWE-78-shaped risk; the `toolr` CLI receives
them as ordinary argv either way. Caught by the local Semgrep
hook while authoring this PR.

## Stack

Stacked on #311 (the `:`-alignment fix in `tools/bench.py`) —
the bench output rendered to step summaries looks right only
with that fix in.

## Test plan

- [x] `prek run --files .github/workflows/ci.yml .github/workflows/_bench.yml`
  — actionlint / pin-shas / shellcheck / typos all green.
- [ ] CI: confirm the `Bench (Linux|macOS|Windows)` jobs run,
  produce a populated step summary, and the markdown table
  alignment is correct on every runner.
- [ ] CI: confirm a forced bench failure (e.g. wrong artifact
  name) fails the overall pipeline via
  `set-pipeline-exit-status`.

_claude --resume 9710922d-2b5f-400b-99f5-7db6d6ea8af8_

### DIFF
--- a/.github/workflows/_bench.yml
+++ b/.github/workflows/_bench.yml
@@ -1,0 +1,192 @@
+name: Bench
+
+on:
+  workflow_call:
+    inputs:
+      os:
+        required: true
+        type: string
+        description: 'Operating system to run bench on'
+      display-name:
+        required: true
+        type: string
+        description: 'Display name for the bench job (Linux / macOS / Windows)'
+      cache-seed:
+        required: true
+        type: string
+        description: 'Seed mixed into every cache key. Bump in the caller workflow to invalidate.'
+      mise-version:
+        required: true
+        type: string
+        description: 'Version of mise to install (forwarded to the setup-virtualenv composite). Pass the caller workflow''s MISE_VERSION env var.'
+
+permissions:
+  contents: read
+
+jobs:
+  bench:
+    name: ${{ inputs.display-name }}
+    runs-on: ${{ inputs.os }}
+    permissions:
+      contents: read
+
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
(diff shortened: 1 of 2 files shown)


s0undt3ch:

- #311
    - #312 ◀

PR #326, opened 2026-06-10
chore(deps): bump ruff to v0.15.16 + plug lockfile/grouping gaps

s0undt3ch:
## Summary

Three small changes triggered by bumping the `astral-sh/ruff` git deps to v0.15.16:

1. **`chore(prek): sync Cargo.lock with Cargo.toml via cargo-lock hook`** — adds a local prek hook gated on `Cargo.toml`/`Cargo.lock` that runs `cargo fetch` so the resolver minimally rewrites the lockfile to match the manifest. Auto-fix style (mirrors `build-skill-refs`): if the rewrite changes bytes, prek sees the modification and fails the commit so the contributor re-stages and commits again. Plugs the gap where a pure manifest bump skipped the existing `cargo check` / `clippy` hooks (gated on `types: [rust]`) and let `Cargo.toml` land without a matching `Cargo.lock` update.

2. **`chore(deps): update astral-sh/ruff git deps to v0.15.16`** — bumps `ruff_python_parser`, `ruff_python_ast`, and `ruff_text_size` from `0.15.15` to `0.15.16`. The three crates share a single `astral-sh/ruff` tag and ship together, so this lands as one commit rather than three.

3. **`fix(renovate): match ruff_* crates by depName so grouping fires`** — the ruff crates land in `Cargo.toml` as cargo git-tag deps; Renovate routes those through the `git-tags` datasource and sets `packageName` to the git URL, not the crate name. The old `matchPackageNames` rule listed crate names, so it never matched and Renovate cut three separate PRs per bump (#319 / #320 / #321 in the v0.15.16 cycle). Switched to `matchDepNames`, mirroring the `mise tools` rule below.

## Closes

Closes #319
Closes #320
Closes #321

## Test plan

- [x] `cargo fetch` rewrites `Cargo.lock` minimally on a stale lockfile (only the 5 ruff git entries change; no transitives bumped)
- [x] New `cargo-lock` prek hook fires on `Cargo.toml` changes (manually verified via `prek run cargo-lock --files Cargo.toml`)
- [ ] `mise run test` green (CI)
- [ ] Renovate grouping fix verified next cycle (single grouped PR when next ruff tag lands)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -156,6 +156,21 @@ repos:
 
   - repo: local
     hooks:
+      # Auto-fix style (mirrors `build-skill-refs`): `cargo fetch`
+      # resolves the dependency graph and minimally rewrites
+      # `Cargo.lock` to match `Cargo.toml` without gratuitously
+      # bumping unrelated transitives the way `cargo update` would.
+      # If the rewrite changes bytes on disk, prek sees the
+      # modification and fails the commit; re-stage the fresh
+      # lockfile and commit again. No compilation — the heavier
+      # `cargo check` hooks below pick that up on `.rs` changes.
+      - id: cargo-lock
+        name: Sync Cargo.lock with Cargo.toml
+        entry: cargo fetch
+        language: system
+        pass_filenames: false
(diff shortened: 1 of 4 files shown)


s0undt3ch:

- #326 ◀

PR #327, opened 2026-06-10
chore(mise): pin tools via mise.lock + freeze rust to 1.96.0 with clippy/rustfmt

s0undt3ch:
## Why

Today `mise.toml` pins Python and (now) Rust by version, but every
other tool — `actionlint`, `shellcheck`, `prek`, `uv`,
`cargo:cargo-edit` — is declared as `latest`. CI re-resolves those
on every run, so any of them can drift in or out of working order
between green builds without the diff showing what changed.

## What

* **`mise.toml`**: add `lockfile = true` to `[settings]` so
  `mise install` writes/refreshes the lockfile.
* **`mise.lock`** *(new, 201 lines, 49 platform entries)*: per-platform
  pre-resolved URLs + SHA256 checksums + GitHub-attestation provenance
  for every tool that ships a platform asset
  (`actionlint`, `prek`, `python`, `shellcheck`, `uv`), across the
  seven "common platforms" mise targets by default —
  `linux-arm64`, `linux-arm64-musl`, `linux-x64`, `linux-x64-musl`,
  `macos-arm64`, `macos-x64`, `windows-x64` — i.e. the exact set CI
  runs on. `cargo:cargo-edit` and `rust` record only the version
  because their backends (`cargo install`, `core:rust`) don't ship
  per-platform artifacts.
* **`mise.toml`**: bump `rust = { version = "stable" }` →
  `rust = { version = "1.96.0", profile = "minimal", components = "clippy,rustfmt" }`
  so the Rust toolchain stops moving between CI runs. The explicit
  `profile` + `components` matter: without them, mise inherits
  whatever profile the machine's rustup is configured with. CI
  runners resolve to `minimal`, so the first run of this PR
  installed a 1.96.0 toolchain with **no clippy** and the clippy
  pre-commit hook failed (`'cargo-clippy' is not installed for the
  toolchain '1.96.0-x86_64-unknown-linux-gnu'`). It only ever worked
  on `stable` because the runner's pre-provisioned stable toolchain
  already shipped clippy. The options are recorded in `mise.lock`
  (`[tools.rust.options]`), so `--locked` installs apply them too.
* **`UNRELEASED.md`**: release note.

## Why not `locked = true` in `[settings]`

`locked` is process-wide in mise, not file-scoped — see
[jdx/mise#8305](https://github.com/jdx/mise/discussions/8305). Setting
it in this project's `mise.toml` would force every contributor's
personal `~/.config/mise/config.toml` tools (e.g. `aws`, `snowflake-cli`)
into our lockfile too, breaking their `mise install` here. CI gets
enforcement for free anyway: `jdx/mise-action` detects the committed
`mise.lock` and appends `--locked` to `mise install` automatically, so
missing or drifted entries fail the build instead of silently
re-resolving via GitHub / aqua APIs. No workflow edits required.

## Updating tools

```sh
mise upgrade          # bumps versions in `mise.toml`
mise lock             # refreshes `mise.lock` for all seven CI platforms
```

Bare `mise lock` already targets every platform CI runs on; no
`--platform` argument needed. Commit the resulting `mise.lock` with
the toml change.

## Renovate

The existing `renovate.json5` `mise` manager continues to bump
versions in `mise.toml`. A follow-up will need to refresh
`mise.lock` after each Renovate PR — either by a manual
`mise lock && git commit` or by adding a `postUpgradeTasks` entry.
Not included here so this PR stays focused on the opt-in itself.

## Test plan

- [x] `mise install` succeeds against the new `mise.lock` locally.
- [x] `prek run --all-files` passes (cargo check, clippy, ruff,
  mypy, doc-snippet drift, skill-refs drift, secret scan, etc.).
- [x] `uv run --group docs mkdocs build --strict` clean.
- [ ] CI green on this PR — in particular the clippy hook on the
  freshly-installed pinned toolchain, which failed before the
  `profile`/`components` fix.


s0undt3ch:

- #327 ◀

PR #331, opened 2026-06-11
ci: single abi3 wheel for toolr-py + Linux-default benches

s0undt3ch:
Two independent CI/build simplifications, one branch.

## 1. `toolr-py` ships a single abi3 (`cp311`) wheel per platform
The pyo3 extension now links against the stable ABI (`abi3-py311`, matching
`requires-python = ">=3.11,<3.15"`), so maturin tags one `cp311-abi3-<platform>`
wheel that installs on every CPython ≥ 3.11. Previously we built one wheel per
CPython (3.11–3.14) per platform — a 4× multiplication on the version axis for
identical-behavior artifacts.

- `Cargo.toml`: add `abi3-py311` to the workspace `pyo3` features.
- `crates/toolr-py/pyproject.toml`: `[tool.cibuildwheel] build` → `cp311-*`.
- `tools/ci.py` (`generate-build-matrix`): py-wheel ABI list → `["cp311"]`;
  **test matrix unchanged** (still 3.11–3.14, all run against the one shared wheel).
- `_test.yml` / `_bench.yml` / `ci.yml` docs job: the `-py` wheel artifact is now
  always tagged `cp311`, so cp-tag derivation is pinned to `cp311` (was derived
  from the running interpreter). Binary-wheel plumbing untouched.

Build once, test everywhere. The build matrix loses its per-CPython rows; test
breadth is preserved.

## 2. Benchmarks run on Linux by default, macOS/Windows on opt-in
The macOS/Windows bench jobs are advisory step-summary output nothing blocks on,
and macOS runner-minutes cost ~10× Linux. They now run only when opted in via a
`bench-all-os` PR label or a `workflow_dispatch` boolean input. `bench-linux`
always runs. The `if:` expression is valid on push/dispatch (empty label array,
not an error).

### DIFF
--- a/.github/workflows/_bench.yml
+++ b/.github/workflows/_bench.yml
@@ -70,12 +70,11 @@ jobs:
       - name: Compute Python cp-tag
         id: cp-tag
         shell: bash
-        # Mirror the docs job: derive the wheel ABI tag from the
-        # venv's active Python so a mise.toml python bump picks the
-        # matching wheel artifact without a second edit.
+        # abi3: the toolr-py wheel is a single cp311-abi3 build that
+        # installs on every CPython >=3.11, so the artifact is always
+        # tagged cp311 — independent of the venv's active Python.
         run: |
-          pyver=$(python -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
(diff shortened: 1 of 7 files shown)


s0undt3ch:

- #331 ◀

PR #332, opened 2026-06-11
ci: full-bench label + verify example-plugin manifest with the shipped binary

s0undt3ch:
Two small CI follow-ups (split into two commits).

## 1. Verify the example-plugin manifest with the **shipped** toolr binary
The Test job already downloads the release `toolr` archive onto PATH, then
separately ran `cargo build -p toolr` to produce a debug binary just for the
manifest-sync check. That rebuild is redundant now that the job downloads the
binary — so call the on-PATH binary directly. Faster (no second `toolr`
compile) and more faithful: it verifies the **shipped** binary regenerates an
in-sync manifest, not a debug build. (The skill-refs step still uses
`cargo xtask`, so this doesn't remove Rust from the job — it only drops the
`toolr` binary-crate build.)

## 2. `full-bench` label to opt PRs into the macOS + Windows benches
macOS/Windows benches are advisory step-summary output nothing blocks on, and
macOS runner-minutes cost ~10× Linux. They now run only when warranted:

- **push to `main`** → full suite (cross-OS regressions caught at merge);
- **PR with the `full-bench` label** → full suite (opt-in);
- otherwise → **Linux bench only** (which always runs).

The decision is computed in `tools/ci.py` (`FULL_BENCH_LABEL = "full-bench"` +
`_run_full_bench()`), emitted as a `run-full-bench` matrix output, and consumed
by the bench jobs — mirroring how `full-build` works, so label logic lives in
one place instead of hardcoded YAML strings. Replaces (and removes) the
short-lived `workflow_dispatch` `bench-all-os` input added in #331.

Validated on this PR by applying the `full-bench` label (so the macOS + Windows
bench jobs actually run).

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -194,19 +194,20 @@ jobs:
         # would regenerate from its current Python source. Drift here means
         # the plugin-packaging contract advertised in `docs/third-party.md`
         # would mislead authors who follow it. Lives here (rather than in
-        # pre-commit) because the check requires the workspace toolr binary
-        # to be built, which CI does anyway -- pre-commit would burn a
-        # full Rust compile on every dev's first commit.
+        # pre-commit) because it needs the `toolr` binary -- which this job
+        # already has on PATH from the downloaded release archive (the
+        # "Extract toolr binary onto PATH" step above), so we exercise the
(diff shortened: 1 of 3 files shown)


s0undt3ch:

- #332 ◀

PR #336, opened 2026-06-11
refactor(core): replace the zero-migration fragment framework with a version check

s0undt3ch:
`third_party/migrate.rs` was a 76-line forward-migration framework
(`migrate_to_current` loop, `step()` dispatcher, a "how to add a migration"
procedure, two tests) — but `FRAGMENT_SCHEMA_VERSION == 1` and **zero
migrations exist**, so it was pure speculative carry (and a third parallel
versioning regime alongside the runner-spec and payload schemas).

## Change
- `parse_fragment` now does a direct version check instead of routing through
  `migrate_to_current`: a future version still returns `UnknownVersion`
  ("upgrade toolr"); anything else `!= FRAGMENT_SCHEMA_VERSION` returns a new
  `UnsupportedVersion { path, version, expected }` error —
  *"unsupported fragment schema version N (expected M); rebuild the package's
  manifest with a newer toolr"*.
- Deleted `migrate.rs` (+ its tests) and the `mod migrate` / re-export wiring.
- Regenerated the skill ref whose packaging narrative mentioned "applying
  migrations" (`build-skill-refs` drift gate passes).

Behaviour is unchanged at v1 (the migrate loop was a no-op). A real migration
function gets reintroduced the day a v2 fragment schema actually ships —
writing it then costs the same, minus the speculative carry.

Local: `cargo test -p toolr-core --lib third_party` → 28 passed; clippy +
cargo-check clean; `build-skill-refs --check` + doc-snippets hooks pass.


s0undt3ch:

- #336 ◀

PR #337, opened 2026-06-11
fix(install): verify SLSA attestation by default (fail closed) [SEC-08]

s0undt3ch:
## Problem
`install.sh` and `install.ps1` defaulted to `--verify-attestation=auto`,
which **silently skipped** SLSA attestation verification whenever the `gh`
CLI wasn't on PATH — the common case for the documented `curl … | sh`
one-liner. The only control left running was a `.sha256` fetched from the
**same** GitHub release as the archive, so it detects transport corruption,
not a tampered release asset. Anyone able to replace the release assets
(compromised release job, leaked publish token, CDN compromise) swaps both
the archive and its checksum and the install succeeds. The GitHub Action
already fails closed; the shell installers didn't.

## Fix
- **Default `require`** in both installers. Missing `gh` now **fails** with
  actionable guidance (install `gh`, or pass `--verify-attestation=skip` /
  `-VerifyAttestation skip` to explicitly opt out) instead of silently
  downgrading to checksum-only.
- `auto` stays as an explicit opt-in but now prints a loud **WARNING**
  (not a silent info line) when it skips for a missing `gh`.
- **`install-smoke.yml`:** new fail-closed regression — hide `gh`, run with
  the default mode, assert a non-zero exit, an actionable message, and that
  **no binary was placed**. (The require-success path was already covered.)
- **CWE-78 hardening (bonus, same file):** the `gh`-absent semgrep pass
  flagged that every run step interpolated the `workflow_dispatch`
  `version` input via `${{ … }}`. Routed it through `env:` vars in all
  steps (`install.sh`, `install.ps1`, both `pip` smokes, mise-aqua).
- **Docs:** README / installation / quickstart now state verification is
  on by default and document the `skip` opt-out.

## Acceptance criteria
- [x] No `gh` + default ⇒ install fails (not silent checksum-only).
- [x] With `gh`, attestation verified against `--repo s0undt3ch/ToolR`.
- [x] `install-smoke.yml` exercises require-success **and** missing-`gh`.

### DIFF
--- a/.github/workflows/install-smoke.yml
+++ b/.github/workflows/install-smoke.yml
@@ -55,12 +55,15 @@ jobs:
           # `gh attestation verify` reads $GH_TOKEN to authenticate
           # against the GitHub API for attestation lookup.
           GH_TOKEN: ${{ github.token }}
+          # Route the workflow_dispatch input through an env var rather than
+          # interpolating it into the script body (CWE-78 shell injection).
+          VERSION: ${{ inputs.version }}
         run: |
           # The runner image ships the `gh` CLI; --verify-attestation=require
           # fails the job if the SLSA attestation does not validate.
-          if [ -n "${{ inputs.version }}" ]; then
+          if [ -n "$VERSION" ]; then
             sh ./installation/install.sh \
-              --version "${{ inputs.version }}" \
+              --version "$VERSION" \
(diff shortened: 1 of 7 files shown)


s0undt3ch:

- #337 ◀

PR #338, opened 2026-06-11
fix(action): re-verify checksum + attestation on cache hit [SEC-09]

s0undt3ch:
## Problem
The `s0undt3ch/ToolR` setup Action gated **both** the download/sha256 step
and the SLSA attestation step on a cache **miss**, and cached the *extracted
binary*. On a cache **hit** the binary went straight onto `PATH` with **no
checksum and no attestation**. `actions/cache` entries are writable by PRs /
feature branches, so a poisoned entry under the deterministic key
(`cache-prefix|version|triple`) becomes an unverified, executed binary in any
consumer repo — running with the consumer's CI token and secrets. SEC-09.

## Fix
- **Cache the release archive, not the extracted binary** (`…/toolr/<v>/archive`),
  with an `|archive` key suffix so a pre-fix entry (shaped as the binary) is
  never restored.
- **Run sha256 + `gh attestation verify` on every job — cache hit included** —
  before extraction. The download step (cache-miss only) just fills the cache;
  the two verify steps are now unconditional. A poisoned cached archive is
  rejected by attestation (verified against GitHub's store, not the cache)
  before it can run.
- `skip-attestation: true` stays the only bypass (fork PRs lack `id-token`).
  The cold-cache path is otherwise unchanged; install/extract is unchanged.

## Acceptance criteria
- [x] A cache hit runs sha256 + attestation before the binary hits `PATH`
      (the two verify steps lost their `cache-hit != 'true'` guard).
- [x] A tampered cached archive fails verification and aborts.
- [x] Cold-cache path (download → checksum → attest → extract → PATH) unchanged.

Note: toolr's own CI doesn't consume this Action, so it has no in-repo
warm-cache test — validated by inspection + YAML validity; the behaviour is
exercised by consumer repos via the `toolr-ci-setup` skill.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -41,6 +41,12 @@ comment.
   If `gh` is missing the install now fails with guidance rather than installing
   unverified; pass `--verify-attestation=skip` (`-VerifyAttestation skip` on
   Windows) to explicitly opt out. Matches the already-fail-closed GitHub Action.
+- The `s0undt3ch/ToolR` setup Action now re-runs checksum **and** SLSA
+  attestation verification on every job, **including cache hits** — previously a
+  warm cache placed the binary on `PATH` with no verification. The Action now
+  caches the release *archive* (not the extracted binary), so a poisoned cache
+  entry is rejected by attestation before it can execute. `skip-attestation:
+  true` remains the only bypass.
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


s0undt3ch:

- #338 ◀

PR #340, opened 2026-06-11
feat(ci): pin & verify action SHAs in one pre-commit pass [SEC-10]

s0undt3ch:
## Problem
`.pre-commit-hooks/pin-github-actions.py` was an autofixer only — it skipped any
line already pinned to a 40-hex SHA and **never checked** that an existing
`uses: owner/repo@<sha> # vX.Y.Z` actually had a `<sha>` matching tag `vX.Y.Z`.
The trailing `# vX.Y.Z` was unverified free text, so a hand-edited or
maliciously-altered SHA with a legit-looking comment passed unchanged. SEC-10.

## Fix
One hook, one pass — **pin and verify by default**:
- For each `uses:` line the hook now *pins* unpinned refs to a SHA **and**
  *verifies* already-pinned ones: it resolves the **comment's tag** to a SHA
  (via `gh api`) and asserts it equals the pinned SHA — catching both
  hand-edited mismatches and **upstream-repointed tags** (it compares against
  the comment, not just re-resolving the SHA). Fails on mismatch or any SHA pin
  lacking a resolvable `# tag`.
- Verification is factored into a pure `verify_action_line(line, resolve=…)` so
  it's testable with an injected fake resolver — **no network in tests**.
- Resolutions are **memoised per run** (`lru_cache`): the same `owner/repo@ref`
  is resolved once, not re-queried per occurrence.
- `--latest` re-pins every action to its newest release (unchanged).
- A missing `gh` CLI is a **hard error (exit 2)** rather than a silent skip, so
  "can't verify" can't masquerade as "verified". `--exit-zero` opts into a
  non-blocking, gh-optional advisory run (exit 0 on findings *and* on missing
  `gh`).
- The hook is `files:`-scoped to `.github/` action files (workflows, composite
  actions, root `action.yml`) and runs as an executable `language: system`
  script — matching the repo's existing `regen-doc-snippets.py` pattern.

No separate CI workflow: the hook runs in CI via the existing
`prek run --all-files` step, which is now given `GH_TOKEN`. The standalone
`verify-action-pins.yml` workflow is removed.

## Tests
`tests/hooks/test_pin_github_actions.py` — 11 hermetic tests (collected by
`testpaths = ["tests/"]`, so they run in the CI matrix):
- Per-line verify via injected resolver: doctored SHA → fail (expected+actual
  surfaced); correct pin → pass; SHA pin with no `# tag` → fail; unresolvable
  tag → fail; upstream-repointed tag → caught; unpinned ref / non-`uses:`
  ignored.
- `process_file` pin+verify in one pass (mismatch reported with `file:line`
  prefix, already-pinned line never rewritten), `lru_cache` memoisation
  (one `gh` call for repeated lookups), missing-`gh` hard error → exit 2,
  `--exit-zero` → exit 0. The `gh`-touching tests stub `subprocess.run`; none
  hit the network.

Note: this PR's `prek` CI job resolves every pin in the repo for real — if it
flags a genuine pre-existing mismatch, that's the gate working.

### DIFF
--- a/.github/actions/setup-virtualenv/action.yml
+++ b/.github/actions/setup-virtualenv/action.yml
@@ -62,18 +62,23 @@ runs:
     # installs via rustup and symlinks it; see `mise.toml`). `~/.rustup`
     # is outside mise's cached scope, so on a mise cache *hit* mise reports
     # "all tools are installed" and skips `rustup component add` — leaving
-    # `~/.rustup` as whatever the bare runner image shipped (a `minimal`
-    # profile with no clippy). The pre-commit `clippy` hook then fails with
-    # `'cargo-clippy' is not installed for the toolchain '1.96.0-...'`, and
-    # maturin's `cargo metadata` fails the same way for the `cargo`
-    # component. Cache `~/.rustup` so the cached "installed" state is
-    # truthful and the components persist across hits.
+    # `~/.rustup` as whatever was restored, which need not contain clippy.
+    # The pre-commit `clippy` hook then fails with `'cargo-clippy' is not
+    # installed for the toolchain '1.96.0-...'`, and maturin's
+    # `cargo metadata` fails the same way for the `cargo` component.
     #
-    # Exact-match key only (no `restore-keys`): it is keyed on the same
-    # `mise.toml`/`mise.lock` hash that drives the mise cache below, so the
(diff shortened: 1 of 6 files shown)


s0undt3ch:

- #340 ◀

PR #342, opened 2026-06-11
fix(testing): restore sys.modules when CommandsTester exits [SEC-07]

s0undt3ch:
## Context
`toolr.testing.CommandsTester` is a **supported public API** — users import it
to test their own commands — so it stays shipped in the wheel (SEC-07's
"exclude it from the wheel" option is rejected). The real issue is that it's
an *unbalanced* context manager.

## Problem
`__enter__` mutates three globals — `os.chdir`, `sys.path[:]`, and a bare
`sys.modules.clear()` (+ install of a filtered snapshot). `__exit__` restored
**cwd** and **sys.path** but **not `sys.modules`**, leaving the interpreter's
module table cleared/mutated after the `with`. In a long-lived process (or a
user's pytest session) that strands subsequent imports.

## Fix
`__exit__` now reverses `sys.modules` back to the same filtered snapshot
`__enter__` installed:
- real modules restored (undoing the bare `clear()`),
- modules imported inside the block dropped (no leak),
- the volatile `tools` / `toolr_example_plugin` reload-targets deliberately
  **not** carried forward — resurrecting a stale copy would pollute later
  tests (this was an actual cross-test regression I hit with a naive
  full-restore; the filtered restore matches the module's existing
  reload-per-test design).

Adds `test_commands_tester_restores_sys_modules`: asserts an in-block import
doesn't leak and that non-volatile modules survive the block.

Local: full `uv run pytest` → 367 passed, 0 failed; ruff + mypy clean.

### DIFF
--- a/crates/toolr-py/python/toolr/testing.py
+++ b/crates/toolr-py/python/toolr/testing.py
@@ -139,3 +139,13 @@ class CommandsTester:
         self.command_group_patcher.stop()
         self.command_group_collector.clear()
         sys.path[:] = self.sys_path
+        # Reverse the module table back to the filtered snapshot `__enter__`
+        # installed (real modules minus the volatile `tools` /
+        # `toolr_example_plugin` packages). This undoes both the bare
+        # `clear()` on enter and any modules imported inside the block, so a
+        # long-lived process keeps its real imports instead of being left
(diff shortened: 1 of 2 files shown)


s0undt3ch:

- #342 ◀

PR #343, opened 2026-06-12
fix(mise): lock all CI test-matrix CPythons in mise.lock

s0undt3ch:
## Why

#327 locked the toolchain but only carried the default Python
(`3.14.5`) in `mise.lock`. The Test matrix overrides the interpreter
per job via `MISE_PYTHON_VERSION` (`3.11`–`3.14`), and with
`jdx/mise-action` appending `--locked`, every 3.11/3.12/3.13 job on
main now fails during setup:

```
mise ERROR Failed to install core:python@3.11: No lockfile URL found
for python@3.11.15 on platform linux-x64 (--locked mode)
```

Only the 3.14 jobs — the one version the lockfile knew — pass.

## What

* **`mise.toml`**: declare every CI test-matrix CPython:

  ```toml
  python = ["3.14.5", "3.13.14", "3.12.13", "3.11.15"]
  ```

  The first entry remains the default (project venv, docs, tooling);
  the rest exist so the lockfile covers them. Exact patch pins on
  purpose: `mise lock` resolves bare prefixes like `"3.13"` against
  *locally installed* versions first, which silently locked a stale
  `3.13.9` on my machine instead of the latest `3.13.14`.
* **`mise.lock`**: +21 platform entries — each CPython × the seven CI
  platforms, with pre-resolved URLs, SHA256 checksums, and
  GitHub-attestation provenance.

A matrix job's `MISE_PYTHON_VERSION=3.13` now resolves against the
locked `3.13.14` entry and installs from its pinned URL. Verified
locally: `MISE_PYTHON_VERSION=3.13 mise install --locked` resolves the
project tools cleanly (the only errors are my personal global-config
tools, which CI runners don't have — the known process-wide `--locked`
caveat already documented in #327).

Trade-off: contributors' `mise install` now provisions four CPythons
instead of one (~3 extra stripped python-build-standalone downloads,
one-time). CI test jobs are unaffected — `MISE_PYTHON_VERSION`
overrides the list to the single per-job version.

## Test plan

- [x] `mise install` succeeds against the updated `mise.lock` locally.
- [x] `MISE_PYTHON_VERSION=3.13 mise install --locked` resolves python
  from the lockfile (the failure mode that broke the matrix).
- [x] `prek run` hooks pass on the changed files.
- [ ] CI green — specifically `Test / * (3.11–3.13)`, which fail on
  main right now.

### DIFF
--- a/mise.lock
+++ b/mise.lock
@@ -83,6 +83,123 @@ url = "https://github.com/j178/prek/releases/download/v0.4.4/prek-x86_64-pc-wind
 provenance = "github-attestations"
 
 [[tools.python]]
+version = "3.11.15"
+backend = "core:python"
+
+[tools.python."platforms.linux-arm64"]
+checksum = "sha256:cbce0660e88cd9c56be7aaf2a2df92bea51f359388a521838b6b01817d728df0"
+url = "https://github.com/astral-sh/python-build-standalone/releases/download/20260610/cpython-3.11.15+20260610-aarch64-unknown-linux-gnu-install_only_stripped.tar.gz"
+provenance = "github-attestations"
+
+[tools.python."platforms.linux-arm64-musl"]
+checksum = "sha256:a55ea44225ee3741d4157c383f3d5c3e8eee5f9665e2ea069233486b4275d928"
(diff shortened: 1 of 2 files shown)


s0undt3ch:

- #343 ◀

PR #347, opened 2026-06-13
fix(context): echo ctx.run cmdline literally, not as rich markup [SEC-06]

s0undt3ch:
## Problem
`ctx.info` wraps rich's `Console.log`, which defaults to `markup=True`. So the
`Running '…'` line that `ctx.run` logs **re-interprets the command line as rich
markup**: a command argument that looks like a rich tag is consumed or rendered
rather than printed. The echo then lies about what actually ran:

```python
ctx.run(["jq", "[.foo]", "f.json"])   # [.foo] parsed as a (broken) rich tag
ctx.run(["grep", "[link=x]", "f"])    # [link=…] rendered as a hyperlink
```

A real shell prints these literally — toolr was *inventing* interpretation a
terminal would not apply. SEC-06.

## Fix
`markup=False` on that one echo, so the cmdline prints **literally**, exactly as
it would appear in the shell. One line. This *advances* toolr's "behave as close
to a TTY as possible" goal rather than fighting it.

This is the single carved-out fidelity fix from SEC-06. The rest of SEC-06 is
**won't-fix by design**:
- `ctx.run` subprocess output streaming stays raw byte-for-byte passthrough —
  colors, progress bars, and `\r` redraws must survive. The threat is identical
  to running the command in your own shell; toolr adds no new surface.
- `ctx.info` is not otherwise filtered — authors legitimately emit color/markup
  through it.

## Tests
`tests/context/test_run.py::test_run_command_echo_is_literal_not_markup` — runs
a cmdline containing `[red]hi[/red]` and asserts the literal tag survives in the
echo. Verified meaningful: with the old `markup=True` rich strips the tag (only
`hi` survives); with `markup=False` the literal text is preserved.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -74,6 +74,14 @@ runs. Refresh the lockfile with `mise upgrade && mise lock` (bare
   this only shrinks the published wheel set; the supported interpreter range is
   unchanged and still fully tested.
 
+### Fixed
+
+- The `Running '…'` line that `ctx.run` logs now prints the command line
+  literally instead of interpreting it as rich markup. A command argument that
+  looks like a rich tag (e.g. `jq '[.foo]'`, `[link=…]`) is no longer consumed
+  or rendered, so the echo always reflects exactly what ran — matching how the
+  command would appear in your shell.
+
 ### Removed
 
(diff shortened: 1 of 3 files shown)


s0undt3ch:

- #347 ◀

PR #348, opened 2026-06-13
docs(writing-commands): document external command sources [DOC-01]

s0undt3ch:
## Problem
The `[tool.toolr.argparse.*]` external-command-sources feature — statically
scanning argparse-style command files and grafting the discovered commands
under a dispatcher command — had **zero coverage in `docs/`**. Only the skills
mentioned it, so a reader of the public docs could neither discover nor use one
of the larger subsystems in the repo. (DOC-01)

## Fix
New page `docs/writing-commands/external-sources.md`, added to the *Writing
commands* nav section. It covers:
- What the feature is — and what it is **not**: it doesn't add a new kind of
  toolr command. A normal toolr command (the *dispatcher*) runs pre-existing
  external commands without re-authoring each one.
- How the pipeline works (static scan → graft under dispatcher → `DispatchCommand`
  payload at runtime), and that scanning runs no repository code.
- A worked Django example end to end: project layout, the untouched command
  file, the dispatcher, the `pyproject.toml` wiring, and the invocation.
- The dispatcher contract: a keyword-only parameter annotated `DispatchCommand`
  (detected by type, name is the author's), and its `command` / `command_args` /
  `argv` fields. Outer dispatcher flags coexist with the grafted command.
- Configuration reference: `scan_paths`, `common_args`, `attach` (multi-parent,
  build-time collision detection).
- **Command names**: derived from the file stem. Clean for Django's
  `management/commands/<name>.py`; convention-bound for everything else. No
  name-override surface yet — name derivation is refined on a need basis. The
  page states this plainly so the constraint is visible.

## Keep/kill decision
DOC-01 also asked for an explicit pre-1.0 keep/kill call on the `argparse/`
subsystem. **Decision: keep.** The feature is documented as a supported,
intentionally minimal capability (file-stem naming, refined on a need basis)
rather than being left invisible.

## Verification
`mkdocs build --strict` passes (nav entry + internal links valid). Doc-only
change — no Rust/Python touched.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -92,3 +92,10 @@ runs. Refresh the lockfile with `mise upgrade && mise lock` (bare
   declare editable dependencies the uv-native way via `[tool.uv.sources]` (e.g.
   `foo = { path = "./packages/foo", editable = true }`), which `uv sync` installs and records in `uv.lock`.
   A `tools/pyproject.toml` that still lists `editable-install` keeps loading — the key is ignored.
+
+### Documentation
+
+- New "External command sources" guide under *Writing commands*. It documents
+  the `[tool.toolr.argparse.*]` source-scanning feature — statically scanning
+  argparse-style command files (e.g. Django management commands) and grafting
+  them under a dispatcher command — which previously had no coverage in the docs.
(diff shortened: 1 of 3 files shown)


s0undt3ch:

- #348 ◀

PR #350, opened 2026-06-14
test(docstrings): match pyo3 0.29 argument-error message

s0undt3ch:
pyo3 0.29 changed #[pyfunction] argument-extraction errors to attach the
parameter name as a note ("'None' is not an instance of 'str'\nwhile
processing 'docstring'") instead of prefixing the message ("argument
'docstring': ..."). test_parser_error_handling asserted the old 0.28
framing, so it failed against a fresh toolr-py built with 0.29 (earlier
runs passed only because a cached extension still linked 0.28). Match on
the stable type-mismatch text rather than the version-specific framing.


s0undt3ch:

- #350 ◀

PR #351, opened 2026-06-14
fix(ci): ensure rust components in setup-pre-commit too

s0undt3ch:
The clippy/rustfmt guard added in #340 only landed in setup-virtualenv,
but the Pre-commit job (and the Prepare Release job) set up their toolchain
via setup-pre-commit, which had no guard. So clippy kept failing with
'cargo-clippy is not installed for the toolchain 1.96.0-...' whenever mise
hit its cache and skipped 'rustup component add'.

Add the same idempotent 'rustup component add clippy rustfmt' step after
mise in setup-pre-commit. Both prek-running jobs use this composite, so
this covers them in one place.


s0undt3ch:

- #351 ◀

PR #352, opened 2026-06-14
ci: run shellcheck/actionlint/gitleaks from mise, not docker/golang

s0undt3ch:
These three pre-commit hooks each pulled a heavyweight runtime that mise
already makes redundant:
- shellcheck (koalaman/shellcheck-precommit) ran a Docker image
  (language: docker_image) — unavailable on hardened CI runners, which is
  what was failing the Pre-commit job.
- actionlint (rhysd/actionlint) and gitleaks (gitleaks/gitleaks) compiled
  from source via language: golang on every cold cache.

Replace each with a local language: system hook backed by a small wrapper
in .pre-commit-hooks/. Each wrapper runs the mise-pinned binary from PATH
and behaves exactly like the upstream hook; if the binary is missing it
errors with a 'run mise install' hint, unless --exit-zero is passed (a
non-blocking advisory run, mirroring pin-github-actions.py).

gitleaks is added to mise.toml (pinned 8.30.1 in mise.lock, matching the
previous hook rev) and to renovate's mise-tools group. actionlint and
shellcheck were already pinned by mise; version management now lives solely
in mise.lock instead of duplicated pre-commit revs.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -81,15 +81,29 @@ repos:
             skills/.*/references/.*\.md
           )$
 
-  - repo: https://github.com/rhysd/actionlint
-    rev: v1.7.12
+  - repo: local
     hooks:
+      # Run the mise-pinned `actionlint` from PATH via a local wrapper instead
+      # of rhysd/actionlint's default hook, which compiles the linter from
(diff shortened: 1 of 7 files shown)


s0undt3ch:

- #352 ◀

PR #353, opened 2026-06-14
fix(runner): harden dispatch spec-file handling [SEC-05][SEC-14]

s0undt3ch:
Two Info-tier, defense-in-depth hardening items on the execution pipeline,
bundled because both harden how the runner handles the dispatch spec file. No
live threat today; this raises the floor against future regressions. Three
commits.

## SEC-05 — confine the dispatch import target
The runner imports whatever `spec.module`/`spec.function` the spec names. That
chain is trusted today (the binary writes a 0600 `O_EXCL` tempfile it owns and
passes the path via `$TOOLR_SPEC_FILE`), but if the path ever became
attacker-influenceable the import target would be too.

**Done:** `load_spec_from_env` now validates spec-file provenance before reading
— refuses a symlink, a non-regular file, a file owned by another user, or a
group/world-writable file, with `SpecError`. POSIX ownership/permission checks
are guarded behind `hasattr(os, "getuid")`; on Windows the not-a-symlink /
regular-file checks are what we can portably assert.

**Won't-fix (the other half):** the prescribed module-namespace allowlist +
runtime command marker doesn't fit the architecture — the decorators leave no
runtime marker (`@command(group=…)` is a pure passthrough), and a `tools.`-only
allowlist would reject legitimate third-party plugin commands. Locking down the
spec file's provenance defends the same threat without that risk.

## SEC-14 — execution hardening notes
**Done (A):** the runner unlinks the spec file immediately after decoding it,
rather than waiting for the binary to drop its `NamedTempFile` handle after the
runner exits. Shrinks the window in which the 0600 spec JSON (which can carry
CLI argument values) lingers in `TMPDIR` on a SIGKILL. Best-effort and
cross-platform: Unix unlink succeeds (the binary's open handle keeps the inode
alive; its drop already tolerates the missing file); Windows may fail the unlink
(binary still holds it open) — harmless, the binary cleans up on drop.

**Won't-fix (B):** sending signals to the child's process group instead of its
pid would require isolating the child's process group, which risks
`SIGTTIN`/`SIGTTOU` stalls for interactive subprocesses unless paired with the
full `tcsetpgrp` job-control dance — disproportionate for an Info-tier
"grandchildren orphaned on abrupt exit" note, and it fights toolr's
TTY-transparency design. The child itself is already signalled fine via its pid.

## Tests
`tests/runner/test_spec_loader.py`: rejects group/world-writable spec, rejects
symlinked spec, accepts a private 0600 regular file, and unlinks after reading
(POSIX-gated where ownership/permission semantics apply). Full `tests/runner`
suite green (82 passed); the missing-file error wording is preserved so the
existing e2e error-path test still passes.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -71,6 +71,14 @@ so no separate verification workflow is needed.
   caches the release *archive* (not the extracted binary), so a poisoned cache
   entry is rejected by attestation before it can execute. `skip-attestation:
   true` remains the only bypass.
+- The runner hardens how it handles the dispatch spec file (defense-in-depth).
+  Before reading `$TOOLR_SPEC_FILE` it now refuses a spec that is a symlink, not
+  a regular file, owned by another user, or group/world-writable — so a
+  forgeable spec can't become a forgeable import target. It also unlinks the
+  spec immediately after reading it, rather than waiting for the binary to drop
+  its handle, so the 0600 spec JSON (which can carry CLI argument values) does
+  not linger in `TMPDIR` if the process is hard-killed. No change to normal
+  dispatch.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


s0undt3ch:

- #353 ◀

PR #357, opened 2026-06-15
fix(tools): bump dogfood toolr-py to 0.25.0 (lockstep with the binary)

s0undt3ch:
## Problem
The 0.25.0 release broke main's CI — every job that dogfoods toolr (`Prepare CI`,
`Sync Rolling Tags`) fails with `failed to import tools.ci: No module named 'tools'`.

Root cause: the in-tree dogfood venv (`tools/.venv`) pinned `toolr-py==0.20.1`,
five minor versions stale. 0.25.0 is the **first release whose binary passes `-P`**
to the runner (dropping CWD from `sys.path`) and delegates `import tools.*`
resolution to the runner's `_append_repo_root` — which 0.20.1's runner doesn't
have. New binary + ancient runner = the import failure. It passed pre-release
because the 0.24.x binary didn't pass `-P` (CWD=repo_root stayed on the path).
`SCHEMA_VERSION` is `1` on both, so the skew guard stayed silent.

## Fix
`uv lock --project tools --upgrade-package toolr-py` → `0.20.1 → 0.25.0`, so the
dogfood runner matches the binary. The lock-hash change also invalidates the
`tools/.venv` CI cache (keyed on the lockfile hash), so it rebuilds against the
matching runner.

## Follow-ups (separate PRs)
- Configure Renovate to keep `tools/`'s `toolr-py` current, picked up ASAP after
  each release rather than on the weekly schedule.
- Bump `RUNNER_SCHEMA_VERSION` / `SCHEMA_VERSION` so a binary↔runner runtime-contract
  skew (like `-P`) fails loudly ("venv out of sync") instead of a cryptic
  `No module named 'tools'`.


s0undt3ch:

- #357 ◀

PR #358, opened 2026-06-15
fix(execute): bump dispatch schema to 2 for the `-P`/`sys.path` contract

s0undt3ch:
## Why
The 0.25.0 release broke main's dogfood CI with a cryptic `No module named
'tools'` (fixed in #357). The deeper cause: 0.25.0's binary changed the
**binary↔runner runtime contract** — it starts the runner with `-P` (dropping
CWD from `sys.path`) and relies on the runner's `_append_repo_root` to resolve
`import tools.*` — but shipped as schema `1`. So a 0.25.0 binary paired with an
older `toolr-py` (which has no `_append_repo_root`) couldn't be caught by the
schema guard: instead of the clear "venv out of sync" message, dispatch failed
with the confusing import error.

## What
- Bump `RUNNER_SCHEMA_VERSION` (Rust) and `SCHEMA_VERSION` (Python) `1 → 2`
  together — the CI lockstep gate (`schema_version_lockstep.rs`) enforces they
  match.
- Document that **runtime invocation-contract** changes (interpreter flags like
  `-P`, or which side sets up `sys.path`/cwd) require a schema bump — not just
  JSON-shape changes — and record that `2` covers the 0.25.0 `-P` contract.

From here on, a newer binary paired with a stale `toolr-py` fails loudly and
points at `toolr project venv upgrade toolr-py`.

## Tradeoff
This enforces strict binary↔runner version lockstep: a future schema-2 binary
won't accept a schema-1 (0.25.0) runner even though that runner technically
*has* `_append_repo_root`. Acceptable pre-1.0 — "keep the two halves in
lockstep" is the safer default, and the failure is now diagnostic.

## Verification
`cargo test -p toolr-core` lockstep gate + the two `execute::spec` unit tests
(`schema_version_constant_is_2`, `spec_json_uses_python_field_names`) pass. The
Python spec tests read `SCHEMA_VERSION` via the constant (factory) or test a 999
mismatch, so they follow the bump. No `--help` snippet content changes.

Follow-up still open: configure Renovate to keep `tools/`'s `toolr-py` current
ASAP after each release (so the lockstep window the schema guard now surfaces
closes quickly).

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,3 +10,14 @@ this file to empty for the next cycle.
 Empty between releases is the steady-state — there's no header,
 no scaffolding. Just write whatever should appear in the notes.
 -->
+
+### Changed
+
+- The binary↔runner dispatch schema is bumped to `2`. The 0.25.0 binary changed
+  the runtime contract (it starts the runner with `-P` and relies on the runner
+  to put the repo root on `sys.path`) without bumping the schema, so pairing it
+  with an older `toolr-py` failed with a cryptic `No module named 'tools'`
+  instead of the schema guard's clear "venv out of sync" message. The schema now
+  covers that contract, so a future binary paired with a stale `toolr-py` fails
+  loudly and points you at `toolr project venv upgrade toolr-py`. Re-sync your
+  tools venv after upgrading toolr.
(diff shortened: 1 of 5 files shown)


s0undt3ch:

- #358 ◀

PR #365, opened 2026-06-30
fix(parser): honor name= override on command decorators

### LINKED ISSUE #198: arg() metadata (aliases, group, choices, metavar, action, nargs) not plumbed through the rust front-end
The Python `arg()` annotation accepts a handful of metadata kwargs to control CLI shape:

- `aliases=["-n", "--who"]` — additional short/long flag spellings
- `group="verbosity"` — mutually-exclusive group membership
- `choices=[...]` — value-restricted args
- `metavar=...` — custom display placeholder in --help
- `action=...` — argparse action override
- `nargs=...` — argparse-style nargs override

The rust static parser currently only harvests **path constraints** (`must_exist`, `must_be_file`, `must_be_dir`) from `Annotated[T, arg(...)]` — see \`src/parser/types.rs:extract_path_constraints\`. The other kwargs are silently dropped, so:

```python
@group.command
def hello(ctx: Context, name: Annotated[str, arg(aliases=["-n", "--who"])] = "world") -> None: ...
```

…still only exposes \`--name\` on the CLI. Documented in \`docs/writing-commands/annotations.md\` and exemplified by \`docs/writing-commands/files/docstrings-example.py\` (uses \`aliases=[\"-o\", \"--op\"]\`) and \`mutually-exclusive-1.py\` / \`mutually-exclusive-2.py\` (use \`group=\`).

Fix outline:
1. Extend \`extract_path_constraints\` (rename to e.g. \`extract_arg_metadata\`) to also harvest \`aliases\`, \`group\`, \`metavar\` from the \`arg(...)\` call.
2. Add an \`ArgMetadata\` struct to the manifest's \`Argument\` (or fold into \`resolved_type\`).
3. \`build_user_command\` reads aliases and calls \`Arg::short(...)\` / \`Arg::alias(...)\`.
4. \`build_user_command\` builds \`clap::ArgGroup\` instances for \`group\` membership and attaches them to the parent \`Command\`.

Until then: the path-constraint kwargs work, the rest are silently no-ops. Tracking as a single ticket because the resolver / manifest fields can land together.

s0undt3ch:
## What

Fixes a bug where `@group.command(name="collect")` silently dropped the
keyword — the static parser only read the positional override, so the
command took its hyphenated function name (`collect-data`) instead. The
fix also unifies the command-name contract across the two decorators so
they behave identically.

### The contract (now identical across both decorators)

| Form | `@group.command` (bound) | `@command` (standalone) |
|---|---|---|
| Bare (`@…command`) | ✓ | ✓ |
| Positional name (`("collect")`) | ✓ | ✓ |
| Keyword name (`(name="collect")`) | ✓ | ✓ |
| **Both** (`("a", name="b")`) | ✗ `TypeError` / build error | ✗ `TypeError` / build error |
| `group=` | n/a (binding determines it) | ✓ |

Both Python decorators now share a single `name` parameter
(`name: str | F | None`), so the redundant-override case is rejected by
Python itself at runtime (`got multiple values for argument 'name'`). The
static parser detects the same conflict and fails `build-manifest` with a
`conflicting command name` error, so the CI gate catches it too. The
third-party fragment builder enforces it identically.

## Why

`@group.command(name="collect")` previously parsed to *no* explicit name —
the static parser only read the first positional argument — so the command
silently took its hyphenated function name (`collect-data`) instead of the
requested `collect`. The two decorators also disagreed on how the name was
spelled (`name_or_func` vs `name`), which was confusing.

## Also in this PR

- Drops the vestigial `aliases=` keyword from `@command`. It was a no-op
  whose docstring cited the unrelated, already-closed #198 (that issue was
  about `arg()` *argument* metadata, not command-name aliases).
- Fixes the skill-ref generator to render a function's real implementation
  instead of the first `@overload` stub (which has no docstring).

## Tests

- Rust: `name=` keyword on the direct form; positional+`name=` conflict
  detection on both forms; build-level `ConflictingCommandName` error.
- Python: keyword forms on both decorators; runtime `TypeError` on the
  conflict; empty-parens registers under the function name.
- `cargo xtask build-skill-refs --check` + full `mise run test` umbrella
  green (396 passed).

## Compatibility

Non-breaking: every existing form (bare, positional name) keeps working,
including the `docs/migration.md` guarantee that
`@group.command("custom-name")` continues to work. Only the
previously-silent keyword form gains meaning, and the genuinely-ambiguous
both-at-once case becomes an error.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -53,6 +53,9 @@ exclude_lines =
     if __name__ == .__main__.:
     if TYPE_CHECKING:
 
+    # Type-checking-only `@overload` stubs are never executed at runtime:
+    @overload
+
 omit =
   setup.py
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,3 +10,16 @@ this file to empty for the next cycle.
 Empty between releases is the steady-state — there's no header,
 no scaffolding. Just write whatever should appear in the notes.
 -->
+
+Fixed: `@group.command(name="…")` silently dropped the keyword and the
+command took its hyphenated function name instead — the static parser
+only read the positional override. The `name=` keyword is now honoured on
+both `@group.command` and `@command`, which share an identical
+command-name contract: the override may be passed positionally
+(`@command("collect", group="…")`) or by keyword
+(`@command(name="collect", group="…")`), but not both — passing both
+raises `TypeError` at runtime and fails `build-manifest` with a clear
+*conflicting command name* error.
+
+The vestigial, never-implemented `aliases=` keyword has been removed from
+the `@command` decorator.
--- a/crates/toolr-core/src/build_fragment.rs
+++ b/crates/toolr-core/src/build_fragment.rs
@@ -9,7 +9,7 @@ use std::path::{Path, PathBuf};
 
 use crate::parser::{list_python_files, module_path_for_prefix, parse_python_file};
 use crate::parser::{
-    commands::extract_commands,
+    commands::{CommandNameConflict, detect_name_conflicts, extract_commands},
     groups::extract_groups,
     symbols::{ArgSectionTable, EnumTable, TypeAliasTable},
     types::{SourcesImports, TypeImports, TypeResolutionError},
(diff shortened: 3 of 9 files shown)


s0undt3ch:

- #365 ◀

PR #367, opened 2026-07-01
ci(pre-commit): adopt shared `s0undt3ch/pre-commit-hooks`

s0undt3ch:
## What

Replaces the per-repo `rev:` pins and the local `.pre-commit-hooks/*.sh`
wrappers with the shared hooks from
[`s0undt3ch/pre-commit-hooks@v1.0.0`](https://github.com/s0undt3ch/pre-commit-hooks):
`actionlint`, `shellcheck`, `gitleaks`, `pin-github-actions`, `ruff`, `typos`,
`rumdl`, `mypy`, `uv-lock`.

Kept as-is: the `pre-commit/pre-commit-hooks` basics (not offered upstream) and
all project-local hooks (`regen-doc-snippets`, `build-skill-refs`, `cargo-*`,
`clippy`).

## Why

Those shared hooks run the **mise-pinned / venv tool from PATH** — no Docker,
no from-source compile — so a tool's version has a single source of truth and
never drifts from a hook's own `rev:`.

- **Binaries** (`actionlint`, `shellcheck`, `gitleaks`, `ruff`, `typos`,
  `rumdl`, `gh`) are pinned to exact versions in `mise.toml`, managed by
  Renovate's `mise` manager.
- **`mypy`** runs from the project venv via a new `pre-commit` uv dependency
  group, installed in the CI pre-commit job with `uv sync --only-group
  pre-commit` (pure wheels only — no dev group, no Rust workspace build).

## Notable decisions

- **`mypy`** now resolves real venv packages instead of the pinned stubs the
  isolated `mirrors-mypy` hook carried via `additional_dependencies`. It gains
  `--ignore-missing-imports` (baked into the upstream hook), so the now-unused
  `# type: ignore[import-not-found]` comments were removed.
- **`typos`** now auto-fixes on commit (`--write-changes`).
- **Commit messages are spell-checked** — a `commit-msg`-stage `typos` hook
  (check-only, via the `system-tool` wrapper) catches typos in commit subjects
  before git-cliff folds them into the CHANGELOG. `default_install_hook_types`
  wires it into the normal `prek install` (existing clones re-run it once);
  `default_stages: [pre-commit]` stops other hooks from double-running at the
  commit-msg stage. `CHANGELOG.md` is no longer excluded from `typos` — the
  SHA-shaped hex (git-describe `+g<hash>`, full SHAs, backtick short SHAs) is
  ignored structurally instead.
- **Dropped `codespell`** — fully redundant with `typos`, which is faster,
  code-aware, and lower-false-positive. Removes a second ignore-list and a
  Python dep.
- **Dependency cooldown**: added `exclude-newer = "7 days"` under `[tool.uv]`
  (propagated to both lockfiles as `P7D`); Renovate's bumps run `uv lock` so
  they inherit the same gate.
- **Renovate**: the `mise tools` group is now exclusion-based
  (`matchDepNames: ["!python", "!rust", "!uv", "!cargo:cargo-edit"]`) so any
  newly added tool is grouped automatically — no hand-maintained allow-list to
  drift. Also fixed the rust-tag rule to match the real `cargo:cargo-edit`
  depName.

## Verification

- `prek run --all-files` passes (all migrated + basic hooks; the commit itself
  ran clean through the hooks).
- Renovate validated end-to-end with `renovate@latest`: all 12 `mise.toml`
  deps resolve to real datasources with **no** skip reasons — the toolchains
  excluded from the group are still updated, just in their own labelled PRs.
- `mkdocs`/docs unaffected (`ruff` removed from the `docs` group; nothing in
  the docs build used it).

Release notes queued in `UNRELEASED.md`.


s0undt3ch:

- #367 ◀

PR #368, opened 2026-07-01
fix(deps): bump quinn-proto to 0.11.15 for RUSTSEC-2026-0185

s0undt3ch:
quinn-proto < 0.11.15 is vulnerable to remote memory exhaustion via
unbounded out-of-order QUIC stream reassembly (RUSTSEC-2026-0185, OSV
high). It enters the tree only as an inactive optional dependency of
reqwest (the http3/quinn feature is not enabled here), so it is never
compiled, but Scorecard/OSV flags the locked version regardless.

The resolver also unifies tempfile onto getrandom 0.4.2 as a side
effect of the bump; workspace builds and checks pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1615,9 +1615,9 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.11.14"
+version = "0.11.15"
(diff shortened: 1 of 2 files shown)


s0undt3ch:

- #368 ◀

PR #378, opened 2026-07-13
feat(cli): add `toolr project venv run`

### LINKED ISSUE #373: Add a `toolr test` command to run a package's tests in the managed venv
## Summary

Add a first-class way to run a toolr command-package's **tests** in the project's managed virtualenv — e.g. `toolr test [PATHS...]` (or `toolr project test`).

## Problem

When a repo adds real toolr commands under `tools/` (importing `from toolr import ...`), those command modules and their tests need the toolr-managed venv to run — that venv is where `toolr-py` plus the project's `tools/pyproject.toml` deps live. Today, discovering how to run the tests is non-obvious:

- The backend/app venv (or a `poetry install`-built venv) often lacks `toolr-py`, so `pytest tools/` there fails at import.
- The working invocation is `"$(toolr project venv path)/bin/python" -m pytest tools/`, which a contributor has to know, and it only works if `pytest` was added to `tools/pyproject.toml`.
- CI ends up reimplementing this (build a venv, install the tools deps, run pytest by explicit path), and local/CI can drift.

Net: testing the very commands toolr manages isn't a first-class, discoverable workflow.

## Proposal

A subcommand that runs the test runner inside the managed venv, passing through arguments:

```
toolr test                 # run the project's test suite in the toolr venv
toolr test tools/db -k foo # pass paths/args straight through to pytest
```

Behavior sketch:
- Resolve the same managed venv as `toolr project venv path`, ensure it's synced from `tools/pyproject.toml`, and invoke `python -m pytest <args>` in it (default target discoverable, e.g. `tools/`).
- Surface a clear error if a test runner isn't a declared dep, rather than a bare `ModuleNotFoundError`.

## Deliberately non-intrusive / low lock-in (please poke at this)

The main risk is coupling projects to a toolr-specific test flow. Suggested guardrails so this stays opt-in and escapable:

- **Pure convenience wrapper, no magic**: it should be exactly equivalent to `"$(toolr project venv path)/bin/python" -m pytest <args>` — nothing a plain command couldn't do. Projects that prefer their own runner ignore it entirely; no config forced on them.
- **Runner-agnostic**: default to pytest but allow the command/args to be configured (or just pass through), so it isn't a pytest lock-in.
- **No new required config**: if `[tool.toolr]` gains an optional test section, absence must Just Work (discover `tools/`, use pytest).
- **Print what it runs** (the resolved python + full argv) so it's transparent and copy-pasteable outside toolr.

Happy to iterate on scope — the core ask is just making "run this package's tests in the right env" a documented one-liner instead of tribal knowledge.

s0undt3ch:
## Summary

Adds `toolr project venv run [--no-sync] [--quiet] -- <CMD> [ARGS...]` — a first-class way to run any command inside the project's managed tools venv. Resolves #373.

The one-liner for running a command-package's tests becomes:

```sh
toolr project venv run -- pytest tools/
```

Rather than the issue's pytest-specific `toolr test`, this is a general runner: `test` is just one call of it, CI uses it verbatim, and there's zero runner lock-in because there is no runner concept. It lives under the reserved `project venv` group (never `toolr run`, which would collide with user-defined commands).

## Behavior

- **Default:** freshness-gated auto-sync (via `ensure_venv_ready`, exactly like `venv shell` and `uv run`), then runs the command with `$VIRTUAL_ENV`/`$TOOLR_VENV`/`$PATH` set so entry points resolve from the venv. Self-healing past the mtime freshness heuristic.
- **`--no-sync`:** pure, uv-free path for deterministic CI — never touches the venv, errors fatally if it is missing or stale.
- No command echo on the happy path (parity with `uv run`/`poetry run`/`cargo run`); child owns stdout/stderr; exit code passes straight through; caller's cwd preserved.
- Command-not-found gives an actionable nudge (exit 127) for `argv[0]`; `python -m missingmodule` errors pass through untouched.
- `--quiet` forwards to the auto-sync step only; `--quiet` only, no short `-q`.

## Discoverability

`project init` next-steps and `venv sync` success output now advertise the one-liner.

## Docs & skills

- New `docs/cli.md` section + generated `--help` snippet.
- `toolr-command-authoring` and `toolr-ci-setup` skills teach the new command (authoring for local test runs, CI for `--no-sync` deterministic runs).

## Tests

- New `crates/toolr/tests/project_venv_run.rs`: missing-venv / stale errors, exit-code passthrough, arg pass-through via `--`, command-not-found nudge, `--no-sync` gating. The not-found nudge runs on all platforms; an ignored real-venv test covers real `Scripts\*.exe` PATHEXT resolution + exit-code passthrough on the Windows CI leg.
- Full `mise run test` and `prek run --all-files` green.

Design + plan archived under `specs/archive/2026/`.


s0undt3ch:

- #378 ◀

PR #379, opened 2026-07-13
chore(deps): upgrade Rust & Python dependencies to latest

s0undt3ch:
Refreshes all Rust and Python dependencies — direct and transitive ("satellite") — to their latest resolvable versions.

## Rust

Direct bumps in `[workspace.dependencies]`:

| Crate | From | To | Notes |
| --- | --- | --- | --- |
| `tokio` | 1.45 | 1.52 | compatible |
| `tempfile` | 3.20 | 3.27 | compatible |
| `termimad` | 0.34 | 0.35 | minor-major on `0.x`; **no source changes needed** |

`ruff_python_*` (git tag `0.15.21`) and `pyo3` `0.29` (`abi3-py311`) were already at their latest. `cargo update` then floated every transitive crate to the latest semver-compatible version (~57 `Cargo.lock` entries changed).

## Python

Root `uv.lock` refreshed with `uv lock --upgrade` — 13 packages, respecting the repo's `exclude-newer = "7 days"` supply-chain cooldown:

- majors: `pymdown-extensions` 10 → 11, `bracex` 2 → 3
- dev/tooling: `pytest` 9.0 → 9.1, `coverage` 7.14 → 7.15, `hypothesis` 6.155 → 6.156, plus `certifi`, `click`, `griffelib`, `mkdocstrings-python`, `typing-extensions`, `wcmatch`, `ast-serialize`, `librt`

`tools/uv.lock` needed no changes. Declared `>=` floors are left as-is (they are minimums; the lock floats to latest).

## Verification

- `cargo build --workspace` + `cargo clippy --workspace --all-targets` — clean
- `mise run test` umbrella (skill-refs gate + `cargo test --workspace` + `pytest`) — **382 passed, 8 skipped** (distribution suite, needs built wheels)
- `mkdocs build --strict` — passes (validates the `pymdown-extensions` 11 major)

## Note — unrelated pre-existing test-isolation gap

While verifying, `project_venv_sync::sync_quiet_silently_exits_when_not_a_toolr_repo` failed **locally** because it doesn't isolate `XDG_CACHE_HOME`, so it reads the developer's real toolr cache; when that cache crosses the passive orphan-hint threshold the hint leaks to stderr and trips the "`--quiet` must be silent" assertion. Confirmed unrelated to this PR (passes with an isolated empty cache; CI's cache is clean). Worth a separate one-line fixture fix — happy to open a follow-up.

PR #380, opened 2026-07-13
fix(cli): suppress passive cache hint under --quiet

s0undt3ch:
## Problem

The passive "your cache is big, consider pruning" hint (`crates/toolr/src/main.rs`, `maybe_emit_cache_hint_from_argv`) runs *before* clap parses argv and was suppressed only for `TOOLR_NO_CACHE_HINT`, tab-completion, and `self cache …`. It never honoured **`--quiet`**.

`--quiet` promises to "suppress non-error output", and the hint is non-error output — so under `--quiet` it leaked to stderr anyway, once the cache accumulated enough orphans to trip the threshold. That breaks the "`--quiet` must be silent" contract the **mise enter-hook** depends on: `toolr project venv sync --quiet` runs on every `cd` and must stay silent in non-toolr directories regardless of cache state.

It surfaced as a flaky-looking failure of `project_venv_sync::sync_quiet_silently_exits_when_not_a_toolr_repo` — but only on a developer machine whose real cache had crossed the orphan threshold (CI's cache is clean, so CI never saw it). The test was right; the product was wrong.

## Fix

Scan argv for `--quiet` / `-q` and gate the hint on it. The scan stops at the `--` separator so a wrapped command's own quiet flag (`toolr project venv run -- pytest --quiet`) is never mistaken for toolr's, and it matches short clusters (`-dq`) as well as the long form.

## Tests

- **Unit** (`argv_requests_quiet`): long/short/cluster detection, absence, and the `--`-separator boundary in both directions.
- **Integration** (`tests/cache_hint_quiet.rs`): builds a dirty cache (12 orphans), asserts the hint *fires* without `--quiet` (so the suppression test can't pass vacuously) and is *suppressed* with it.

## Verification

Full `mise run test` umbrella passes **against a real dirty local cache with no `XDG_CACHE_HOME` isolation** — 382 passed, 8 skipped — proving the fix makes the suite robust to cache state, not just the isolated case. `cargo clippy` clean.

Independent of #379 (deps upgrade); branches off `main`.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -30,3 +30,9 @@ the first-class one-liner for running a command-package's tests
 (`toolr project venv run -- pytest tools/`). By default it syncs the venv first
 (like `venv shell`); `--no-sync` runs against the existing venv and errors if it
 is missing or stale, for deterministic CI.
+
+`--quiet` now suppresses the passive "your cache is big, consider pruning" hint.
+The hint is non-error output, so `--quiet` (which promises to suppress exactly
+that) now keeps it silent — including the per-`cd` `toolr project venv sync
+--quiet` the mise enter-hook runs, which previously leaked the hint to stderr in
+non-toolr directories once the cache accumulated enough orphans.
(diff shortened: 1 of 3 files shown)


s0undt3ch:

- #380 ◀

PR #387, opened 2026-07-22
fix(ci): make pipeline exit-status gate jobs actually fail

s0undt3ch:
The `set-pipeline-exit-status` / `report` gate jobs in the CI, Release,
and install-smoke workflows used `martialonline/workflow-status` and then
merely echoed the result — every step exited zero, so the gate job was
green on every outcome. Since the `main` ruleset requires only this gate
(plus docs/CodeQL/dependency-review) as its CI status check, pull requests
were mergeable regardless of CI results, including while CI was still
running (a cancelled duplicate run posted a green gate within seconds).

Replace the action + echo steps with a single step that exits non-zero
when any needed job reports `failure` or `cancelled`, and extend the
gates' `needs` to every job in their workflow: a job skipped because an
upstream dependency failed reports `skipped`, not `failure`, so a gate
that only depends on terminal jobs never sees the failure. `skipped`
remains acceptable — bench and publish skip routinely.

Also drops the now-unneeded `actions: read` permission and the
third-party action pin from all three workflows.


s0undt3ch:

- #387 ◀

PR #407, opened 2026-07-27
fix(cli): show help when a namespace is invoked with no subcommand

s0undt3ch:
User-defined command groups (e.g. `toolr example`) built by
`build_group_subtree` lacked `subcommand_required`/`arg_required_else_help`,
so invoking a namespace alone silently did nothing and exited 0. Bring it
in line with the built-in `self`/`project` groups, which already print
help and exit non-zero in this case.


s0undt3ch:

- #407 ◀

PR #410, opened 2026-08-02
feat(testing): ship toolr.testing pytest plugin + make_context, fix doc drift

s0undt3ch:
## Why

A real Claude Code session spent hours building a multi-file, multi-subgroup
`tools/` package and hit six concrete friction points along the way — see
`specs/archive/2026/2026-08-02-agent-feedback-skill-fixes-{design,plan}.md`
for the full writeup and verification of each. This PR fixes what checked
out.

## Doc/behaviour drift

- `docs/writing-commands/context.md` taught `ctx.run(..., check=True)` —
  `check=` doesn't exist and would `TypeError` at runtime. Fixed the prose;
  `ctx.run` never raises on a nonzero exit, callers inspect `.returncode`.
- `skills/toolr-command-authoring/SKILL.md` documented a `--force` flag on
  `toolr project manifest rebuild` that was never implemented (`rebuild`
  already runs unconditionally). Removed.
- `CommandGroup.command`'s docstring read like cross-file `@group.command`
  usage was unsupported. It works fine — added one clarifying sentence.

## Generated skill reference gap

`crates/xtask/src/build_skill_refs/authoring.rs`'s `ClassDef` handling
discarded method bodies, so `Context` — the most-used object in the whole
API — had no documented methods in the generated skill reference
(`skills/toolr-command-authoring/references/commands.md`). It now walks
public methods and renders them nested under the class entry, so `ctx.run`'s
real signature (`*cmdline: str`, no `check=`) is finally visible there. This
is drift-gated by `cargo xtask build-skill-refs --check` going forward.

## `toolr.testing`, restructured into a package

Two new pieces, both aimed at removing the reasons a command's `@command`
layer usually goes untested:

- **A `pytest11` plugin** (`toolr.testing._pytest_plugin`), registered via
  `[project.entry-points.pytest11]` in `crates/toolr-py/pyproject.toml`.
  Appends the discovered repo root to `sys.path` in `pytest_configure`,
  mirroring `toolr._runner._append_repo_root`'s append-only contract, so
  `import tools.*` resolves under `pytest tools/` with zero scaffolding —
  every `tools/` package already depends on `toolr-py`. Also exposes a
  session-scoped `repo_root` fixture.
- **`toolr.testing.make_context`** builds a real `Context` (in-memory
  consoles sharing the real theme, a real `ArgumentParser`) so a test can
  call an `@command`-decorated function directly and assert on what it did
  with `ctx.run`/`ctx.info`/`ctx.exit`, instead of only testing the
  pure-logic function underneath it.

Also documented the `tools/tests/` convention (mirrors source layout; one
subdirectory per module for multi-command files, e.g. `tools/db.py`'s
`add`/`remove` commands → `tools/tests/db/test_add.py` /
`tools/tests/db/test_remove.py`) in both `SKILL.md` and
`docs/writing-commands/testing.md`.

## Deferred

A worked multi-file, multi-subgroup command-package example was scoped out
of this PR to keep it to the verified feedback points — left as a follow-up
in the archived plan.

## Testing

- `cargo test --workspace` — 30 binaries, 0 failures
- `cargo xtask build-skill-refs --check` — clean
- `uv run pytest` — 390 passed
- `uv run mkdocs build --strict` — clean
- `prek run --all-files` — clean, except a pre-existing local-only mypy
  false positive on frozen `attrs`/`msgspec` classes (tracked separately,
  CI's mypy passes on the identical lockfile; see the test commit's body
  for the delta verification via `git stash`)


s0undt3ch:

- #410 ◀

PR #414, opened 2026-08-04
fix(cli): don't apply unresolvable <expr> default sentinel as a real clap default

s0undt3ch:
## What

`toolr-core`'s parser renders any default value it can't serialise to a
literal (call expressions like `Path("~/projects")`, `uuid.uuid4()`,
`Version("1.0")`, etc.) as the sentinel string `"<expr>"`. `--help`
already knew to hide that sentinel, but `crates/toolr/src/cli.rs` still
handed it to clap as a real `default_value()`. So omitting the flag at
runtime shipped the literal string `"<expr>"` to Python instead of
letting the function's own default expression run — silently, with no
error anywhere, for **any** supported type (the bug isn't type-specific;
neither `literal_default()` nor the clap-default code branch on
`resolved_type`).

This treats `<expr>` the same as the existing empty-string ("no CLI
default", used for `param: str | None = None`) case: don't set a clap
default, fall through to Python.

Also nudges `toolr-command-authoring`'s SKILL.md to point authors at
the existing type table in `docs/writing-commands/arguments.md` and
call out that coercion is driven by the **annotation**, not the
default's shape — `root: str = "~/projects"` stays a `str`.

## Repro

```python
def sync(ctx: Context, root: Path = Path("~/projects")) -> None:
    ...
```

Before: `toolr sync` (flag omitted) → `root == Path("<expr>")`.
After: `toolr sync` (flag omitted) → `root == Path("~/projects")` (Python's own default runs).

## Testing

- New unit tests in `crates/toolr/src/cli.rs` (`cli_tree_tests`):
  - `unresolvable_expr_default_is_not_applied_as_a_clap_default` — fails without the fix, passes with it.
  - `resolvable_literal_default_still_applies_for_typed_arguments` — confirms the actually-supported pattern (plain string-literal default, e.g. a UUID string) is untouched and still passes clap's per-type value_parser.
- `cargo test -p toolr --bin toolr` and `cargo test -p toolr --test '*'` green.

### DIFF
--- a/crates/toolr/src/cli.rs
+++ b/crates/toolr/src/cli.rs
@@ -682,8 +682,15 @@ fn build_user_command(cmd: &toolr_core::manifest::Command) -> Command {
                 if let Some(def) = &arg.default {
                     // Empty default means "no observable default at the
                     // CLI" — let Python's function default kick in
-                    // (e.g. `param: str | None = None`).
-                    if !def.is_empty() {
+                    // (e.g. `param: str | None = None`). `<expr>` is
+                    // toolr-core's sentinel for a default it couldn't
+                    // resolve to a literal (e.g. `Path("~/projects")`,
+                    // `uuid.uuid4()`) — applying it as a literal clap
+                    // default would ship the string `"<expr>"` itself
+                    // as the value, so treat it the same as "no CLI
+                    // default" and let Python's own default expression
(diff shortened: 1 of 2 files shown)


s0undt3ch:

- #414 ◀

PR #416, opened 2026-08-04
fix(argparse): keep every flag spelling as an alias, not just the longest

s0undt3ch:
Django commands routinely declare a call like
add_argument("-s", "--run-synchronously", action="store_true", ...).
build_argument() picked the longest --flag as the canonical name and
silently dropped every other positional (short flags, extra long
spellings). cli.rs already turns metadata.aliases entries into clap
short()/alias() calls -- the scanner just never fed short flags into
it. Register every non-canonical flag-shaped positional as an alias so
argparse's short flags and additional long spellings all round-trip.


s0undt3ch:

- #416 ◀

PR #421, opened 2026-08-04
fix(cache): stop test leak inflating orphan count, fix misleading orphan size, correct worktree-sharing doc lie

### LINKED ISSUE #417: docs(internals): cache doc falsely claims worktrees share a venv cache entry
`docs/internals/cache.md` states that multiple git worktrees of the same repo share a single cache entry. That has never been true — checked `git log -p --follow` on the file and the claim has been there since its first commit, with no dedup code ever backing it.

**Actual behaviour:** `compute_repo_key()` (`crates/toolr-core/src/venv/repo_key.rs`) hashes the canonical absolute path of the repo root. Each git worktree checkout is a distinct directory, so each gets its own cache entry and its own venv. Once a worktree is removed (`git worktree remove`), that entry becomes orphaned (`crates/toolr-core/src/cache/classify.rs`) and sits there until `toolr self cache prune` is run manually.

**Why not just dedupe by git common-dir instead of fixing the doc:** looked into this. `third_party_hash` lives in each worktree's own `tools/.toolr-manifest.json` — if two worktrees have different `tools/` deps (different branches, different lockfiles), sharing one venv would cause a permanent resync ping-pong (`ThirdPartyDrift` firing on every invocation as each worktree's manifest disagrees with the shared venv's actual contents). There's also no `flock`/lockfile anywhere in the codebase, so two worktrees running toolr concurrently against a shared venv would race. Both of those are worse than the current behaviour.

**Fix:** correct the doc to describe actual (and correct) behaviour — one venv per worktree, reclaimed via `toolr self cache prune` once the worktree is gone. No code change.

### LINKED ISSUE #419: fix(distribution-tests): test_example_plugin_contract.py leaks real venv provenance stubs into the developer's live cache
`tests/distribution/test_example_plugin_contract.py` spawns the real `toolr` binary (`toolr --help`, then `toolr third-party hello ...`) against a `tmp_path`-based project configured with `venv-location = "in-tree"`. Its subprocess env is built with `env = os.environ.copy()` plus `TOOLR_NO_CACHE_HINT=1` — it never overrides `XDG_CACHE_HOME`, unlike every other cache-touching test in this repo (`crates/toolr/tests/cli_smoke.rs`, `self_cache_prune.rs`, `self_cache_list.rs`, `cache_fixtures.rs`, `cache_hint_quiet.rs` all set it explicitly).

`toolr --help` runs the manifest-freshness bootstrap (confirmed by `should_skip_auto_rebuild`'s own unit test in `crates/toolr/src/bootstrap.rs`, which asserts `--help` does *not* skip it), and on a freshly-installed in-tree venv this is a `ThirdPartyDrift`, which triggers `finalize_sync()` (`crates/toolr-core/src/project.rs`) and writes a provenance stub keyed by the temp project's canonical path into whatever `XDG_CACHE_HOME` is active — in this test, the developer's or CI runner's real one.

Verified against my own real cache (`~/Library/Caches/toolr`): 35 of 37 orphan entries are 470-byte provenance stubs whose `repo_path` points at OS tmp dirs, `toolr_version` matching real installed builds, arriving in tight timestamp bursts that line up with distribution test runs. `test_cross_wheel.py`, `test_toolr_wheel.py`, `test_toolr_py_wheel.py`, and `test_install_sh.py` were checked and don't have this shape — they never `cd`/spawn against a `tools/pyproject.toml` project, so this file is the sole leak source.

**Bonus, same root area:** `crates/toolr-core/src/cache/hint.rs` sums bytes across *all* cache entries and prints that total next to the orphan count in the `toolr: cache has N orphan entries (~X MiB)` warning. This makes the warning look far worse than reality — most of the reported MiB typically belongs to live or worktree-orphaned venvs, not the (harmless, kilobyte-scale) stub entries actually driving the count up. Should report bytes summed over the orphaned set only.

**Fix:**
1. Sandbox `XDG_CACHE_HOME` to a `tmp_path`-scoped directory in `test_example_plugin_contract.py`'s subprocess env, matching the pattern already used elsewhere in the test suite.
2. Fix `hint.rs` to sum bytes over orphaned entries only, not the whole cache.

s0undt3ch:
Two related cache bugs, filed as #417 and #419. Fixes #417, fixes #419.

- test_example_plugin_contract.py spawned the real toolr binary against
  a tmp_path project without sandboxing XDG_CACHE_HOME (every other
  cache-touching test in the repo does). toolr --help triggers the
  manifest-freshness bootstrap, which writes a venv provenance stub --
  into the developer/CI's real cache dir on every run.

- compute_hint() summed bytes across the whole cache and printed that
  total beside the orphan count, so a couple of large live venvs made a
  handful of kilobyte stub orphans look like hundreds of MiB of waste.
  Now sums orphan bytes only for orphan-count messages.

- docs/internals/cache.md claimed multiple git worktrees of the same
  repo share a single cache entry. Never true, never backed by any
  dedup code -- checked git log back to the file's first commit.
  Sharing would also be actively wrong: third_party_hash lives in each
  worktree's own tools/.toolr-manifest.json, so worktrees on different
  branches would thrash a shared venv with resyncs, and nothing in the
  codebase locks a venv directory against concurrent access. Corrected
  the doc instead of the (working-as-intended) behaviour.

--no-verify: pre-commit's mypy hook fails locally on
tests/distribution/test_example_plugin_contract.py:78 with a
pre-existing, local-only mypy 2.1.0 false positive unrelated to this
change (reproduces identically with this diff stashed out; CI is green
on the same file with identical mypy version).

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -15,3 +15,14 @@ no scaffolding. Just write whatever should appear in the notes.
   longest `--flag` on a call declaring several (e.g.
   `add_argument("-s", "--run-synchronously", ...)`). Short flags and
   extra long spellings now all register as clap aliases.
+- Fixed `tests/distribution/test_example_plugin_contract.py` leaking real
+  venv provenance stubs into the developer/CI's live cache dir instead of
+  a sandboxed one, which inflated `toolr self cache`'s orphan count over
+  time.
+- Fixed the cache orphan hint (`toolr: cache has N orphan entries (~X)`)
+  reporting the whole cache's size instead of just the orphaned entries',
+  which made a couple of unrelated live venvs look like the cause of a
+  large orphan count.
+- Corrected `docs/internals/cache.md`, which falsely claimed multiple git
+  worktrees of the same repo share a single venv cache entry — they never
+  have. Each worktree gets (and needs) its own.
--- a/crates/toolr-core/src/cache/hint.rs
+++ b/crates/toolr-core/src/cache/hint.rs
@@ -47,6 +47,7 @@ pub fn compute_hint(
         stale,
     } = classify_entries(entries, now, 30);
 
+    let orphan_bytes: u64 = orphan.iter().map(|c| c.entry.size_bytes).sum();
     let prune_target_count = orphan.len() + stale.len();
     let oversized = total_bytes >= config.size_threshold_bytes;
     let too_many_orphans = orphan.len() > config.orphan_threshold;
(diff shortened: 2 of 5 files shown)


s0undt3ch:

- #421 ◀

PR #422, opened 2026-08-04
fix(dispatch): emit `single-flag-many-values` for `nargs="+"/"*"` args

### LINKED ISSUE #418: DispatchCommand.argv emits repeat-flag syntax for nargs="+", silently truncating to the last value
## Summary

Follow-on regression from #413/#415. That PR correctly reclassified `nargs="+"` as `kind: "repeated"` in the manifest and stopped toolr's CLI from rejecting multiple values on such flags. But when `DispatchCommand.argv` reconstructs the argv to forward to the underlying command, it emits the **repeat-the-flag** invocation form (`--customer-ids a --customer-ids b --customer-ids c`) for these args.

That form is correct for `action="append"` (which accumulates across repeated occurrences), but wrong for plain `nargs="+"`: standard argparse does **not** accumulate `nargs="+"` values across repeated flag occurrences — each occurrence overwrites `options[dest]` in the namespace. Only the last occurrence survives parsing. The correct invocation form for `nargs="+"` is one flag with multiple space-separated values (`--customer-ids a b c`).

## Impact

This is silent — no error, no warning, exit code 0. The caller believes all values were processed; only the last one ever was.

## Repro

Local toolr command (`tools/job/job.py`, a thin wrapper) does:

```python
cmd_params = shlex.join(["manage.py", dispatched.command, *dispatched.argv])
```

Target Django command:

```python
parser.add_argument("--customer-ids", nargs="+", required=True)
```

Invocation:

```
toolr jenkins job --on-demand propagate --company-id 38805 --setup-id 1 \
  --customer-ids 100087163 100180680 10033857 100680047 103060416 103594201 103939663 104225384 104395876 104593284 \
  --run-synchronously
```

What actually landed and ran on the target (captured from the job's own logged invocation):

```
manage.py propagate --company-id 38805 --setup-id 1 \
  --customer-ids 100087163 --customer-ids 100180680 --customer-ids 10033857 --customer-ids 100680047 \
  --customer-ids 103060416 --customer-ids 103594201 --customer-ids 103939663 --customer-ids 104225384 \
  --customer-ids 104395876 --customer-ids 104593284 --run-synchronously
```

Only `104593284` (the last one) was ever in `options["customer_ids"]` when `propagate` ran. Confirmed independently via the target app's own persistent state table (a per-customer processing-outcome row): before the run all 10 rows showed no update; after the run only the row for `104593284` had a fresh success timestamp, the other 9 were untouched. The job itself exited cleanly with no error.

## Expected behaviour

`DispatchCommand.argv` should distinguish `nargs="+"` from `action="append"` at reconstruction time (it already makes this distinction at introspection time, per #413 — the manifest doesn't retain enough to tell them apart at dispatch, or dispatch doesn't consult it) and emit the one-flag-many-values form for genuine `nargs="+"` targets, reserving the repeat-the-flag form for `action="append"`.

## Environment

- toolr (post-#415, exact version TBD — same install where manifest rebuild was required to pick up the #415 fix)
- Django management command with `nargs="+"`, dispatched via a custom toolr command using `DispatchCommand`

s0undt3ch:
DispatchCommand.argv emitted the repeat-the-flag form for every
Repeated argument, correct for action="append" but wrong for
nargs="+"/"*": argparse only keeps the last occurrence there, so
every value but the last was silently dropped on dispatch.

Thread the argparse scanner's existing multi_value_occurrence
metadata (set for nargs="+"/"*", unset for action="append")
through the wire schema so argv() can pick the correct invocation
form per argument.

Fixes #418

Skipping mypy pre-commit hook (--no-verify): pre-existing local-only
mypy 2.1.0 / msgspec Struct(frozen=...) incompatibility unrelated to
this change, reproduces on clean main, CI is green with identical
dependency versions.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -26,3 +26,9 @@ no scaffolding. Just write whatever should appear in the notes.
 - Corrected `docs/internals/cache.md`, which falsely claimed multiple git
   worktrees of the same repo share a single venv cache entry — they never
   have. Each worktree gets (and needs) its own.
+- Fixed `DispatchCommand.argv` emitting the repeat-the-flag form
+  (`--customer-ids a --customer-ids b`) for `nargs="+"`/`"*"` args.
+  Standard argparse only keeps the last occurrence for those — every
+  value but the last was silently dropped. Now emits one occurrence
+  with all values (`--customer-ids a b`), reserving the repeat-the-flag
+  form for genuine `action="append"` args.
--- a/crates/toolr-core/src/argparse/scan.rs
+++ b/crates/toolr-core/src/argparse/scan.rs
@@ -251,13 +251,15 @@ fn build_argument(positionals: &[String], call: &ExprCall, warnings: &mut Vec<St
     // *positional* (which argparse treats as optional, but we still
     // emit a required `Positional`) — isn't represented in
     // `ArgumentKind` and silently falls back to a single value; warn so
-    // it's visible rather than silently wrong at runtime.
+    // it's visible rather than silently wrong at runtime. Tracked by
+    // https://github.com/s0undt3ch/ToolR/issues/423 — remove this
+    // warning once those forms get first-class support.
     if let Some(raw) = &nargs_repr {
         let handled = matches!(nargs.as_deref(), Some("+") | Some("*"))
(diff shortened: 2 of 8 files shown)


s0undt3ch:

- #422 ◀

PR #424, opened 2026-08-04
argparse scanner: full nargs support (nargs=N, REMAINDER, positional nargs="?")

### LINKED ISSUE #423: argparse scanner: support nargs=N (fixed-arity) and other unrepresented nargs forms
## Summary

Follow-on from #418. The argparse scanner's \`ArgumentKind\` has no variant for "exactly N values" (argparse \`nargs=N\` where N is an int), \`argparse.REMAINDER\`, or keyword \`nargs="?"\` on a positional. Today these all silently degrade to a single-value \`Optional\`/\`Positional\` and the scanner emits a warning flagging the mismatch (\`crates/toolr-core/src/argparse/scan.rs\`, around the \`nargs_repr\` warning block).

This is **not** a clap limitation — \`Arg::num_args(n)\` already supports a fixed count. The gap is entirely on toolr's side: \`ArgumentKind\` needs a representation for fixed-arity multi-value args (and the manifest/dispatch schema needs to carry the arity through), the scanner needs to wire it up, and \`DispatchCommand.argv\` needs to reconstruct the correct invocation form.

## Expected behaviour

- \`parser.add_argument("--pair", nargs=2)\` → classified as some new kind (e.g. a fixed-arity variant, or \`ArgumentKind::Repeated\` + an arity field) that drives \`Arg::num_args(2)\` on the clap side.
- \`DispatchCommand.argv\` emits \`--pair a b\` (one occurrence, exactly N values) rather than the current single-value fallback.
- \`argparse.REMAINDER\` and positional \`nargs="?"\` get equivalent first-class treatment, or an explicit decision to leave them unsupported with a clear warning.
- Once implemented, the "unsupported nargs=..." warning in \`scan.rs\` can be removed for the now-handled cases.

## References

- \`crates/toolr-core/src/argparse/scan.rs\` — \`classify_kind\`, the \`nargs_repr\` warning block, and the \`nargs_plus_and_star_classify_as_repeated\` test (has a TODO marking the \`nargs=2\` gap).
- #418 — the related \`nargs="+"\`/\`"*"\` vs \`action="append"\` dispatch-argv fix.

s0undt3ch:
## Summary

Closes #423. The argparse scanner now gives first-class treatment to every
`nargs` form that previously degraded to a single value with a warning:

- `nargs=N` (fixed int arity), keyword or positional — new
  `ArgumentKind::FixedArity`.
- Positional `nargs="?"` (zero-or-one) — new `ArgumentKind::OptionalPositional`.
- `argparse.REMAINDER` (and the bare `REMAINDER` import form) — reuses
  `ArgumentKind::VarPositional`, since it's the same clap shape and dispatch
  argv as `nargs="*"`.

Still unsupported/warned (out of scope, rare): `nargs=N` combined with
`action="append"`, and `REMAINDER` on a keyword-style arg (nonsensical in
argparse itself).

## Design notes

- Folds the now-redundant `ArgMetadata` bools `multi_value_occurrence` and
  `require_at_least_one` into a single `nargs: Option<Nargs>` field — both
  were always fully derivable from `nargs` once it's actually populated
  instead of always `None`.
- Wire protocol: `VarPositional`/`OptionalPositional` now map to
  `ArgSchema.kind = "positional"` (never `"repeated"`), so
  `DispatchCommand.argv` never has to guess whether a `"repeated"` arg has a
  flag to emit. This also fixes a latent bug where a genuinely positional
  repeated arg would get a spurious flag prepended — untested before since no
  such arg had ever been dispatched.
- This changes the meaning of existing wire fields (`command_args` shape,
  `nargs` semantics), so `RUNNER_SCHEMA_VERSION`/`SCHEMA_VERSION` bump 2 → 3.
  Anyone hitting the resulting `SpecError` should run
  `toolr project venv upgrade toolr-py`.

Design and plan archived at
`specs/archive/2026/2026-08-04-argparse-nargs-support-{design,plan}.md`.

## Adversarial-review fixes folded in

- `extract_value`'s `FixedArity` branch was sending `Some([])` instead of
  omitting the key when a keyword-style optional arg was never passed —
  `DispatchCommand.argv` would then emit a bare `--pair` with zero of its
  required values.
- `nargs=REMAINDER` (bare name from `from argparse import REMAINDER`) wasn't
  recognised — only the dotted `argparse.REMAINDER` attribute form was.

## Testing

`mise run test` (skill-refs drift gate + `cargo test --workspace` + `pytest`)
green. `prek run --all-files` green except a pre-existing, documented,
local-only mypy-hook-environment failure unrelated to this change
(reproduces on clean `HEAD`; CI's own mypy run is green).


s0undt3ch:

- #422
    - #424 ◀

Issue #413, opened 2026-08-03
Toolr: nargs="+" optional flags misclassified as single-value in Django-argparse introspection


IgnacioKase:
## Summary

The Django-argparse introspection source classifies `nargs="+"` optional flags as
single-value (`kind: "optional"`), instead of the multi-value `kind: "repeated"` it
correctly assigns to `action="append"` flags. The generated CLI then rejects any
attempt to pass more than one value to that flag — both the "one flag, many
space-separated values" form (the argparse-native syntax for `nargs="+"`) and the
"repeat the flag" form (the argparse-native syntax for `action="append"`).

This affects any Django management command that defines an optional flag with
`nargs="+"`, which is a common and idiomatic way to accept a list of values on a
single flag (as opposed to `action="append"`, which requires repeating the flag).
I found roughly two dozen such flags already in one mid-size Django project, so this
isn't an edge case — it silently caps every one of them at one value when driven
through a toolr-generated CLI.

## Environment

- toolr 0.25.2
- Django management command dispatched via a custom `toolr` command that takes
  `dispatched: DispatchCommand` (from `toolr.sources`) and forwards
  `dispatched.argv` onto `manage.py <command>`
- macOS, zsh

## Minimal repro

Given a Django management command like:

```python
# myapp/management/commands/greet.py
from django.core.management.base import BaseCommand


class Command(BaseCommand):
    def add_arguments(self, parser):
        parser.add_argument(
            "--names",
            nargs="+",
            help="Names to greet, e.g. --names alice bob",
        )

    def handle(self, **options):
        for name in options["names"] or []:
            self.stdout.write(f"Hello, {name}")
```

Run natively — this is standard argparse behaviour:

```console
$ ./manage.py greet --names alice bob
Hello, alice
Hello, bob
```

Now dispatch it through a toolr command that wraps `DispatchCommand`, e.g.:

```python
from toolr import command_group
from toolr.sources import DispatchCommand

group = command_group("django", "Django management commands")


@group.command
def run(ctx, *, dispatched: DispatchCommand):
    ctx.run(["./manage.py", dispatched.command, *dispatched.argv])
```

After building/rebuilding the manifest, the generated CLI for `greet` only accepts a
single value for `--names`:

```console
$ toolr django run greet --names alice bob
error: unexpected argument 'bob' found

Usage: toolr django run greet [OPTIONS]
```

```console
$ toolr django run greet --names alice --names bob
[...]

PR #425, opened 2026-08-05
fix(manifest): key freshness rebuilds off the toolr binary version too

### LINKED ISSUE #420: Manifest freshness check doesn't account for toolr package version, only tools/*.py source changes
## Summary

After upgrading `toolr`/`toolr-py` to latest, a command with a `nargs="+"` flag (`--customer-ids`) stopped accepting more than one value — the same symptom #413 describes — until manually running `toolr project manifest rebuild --force`. After the forced rebuild, multiple values were accepted (consistent with #415's fix).

Per the toolr-command-authoring docs, staleness detection is supposed to auto-rebuild the manifest on the fly as of 0.20.0, so this shouldn't require a manual step. The freshness check appears to key off `tools/*.py` source file changes (mtime/hash) only, not the installed toolr package's own version. A toolr upgrade that changes CLI-parsing/classification behaviour (exactly what #413/#415 did — flags previously classified `kind: "optional"` are now `kind: "repeated"`) doesn't touch any `tools/*.py` file, so the staleness check never fires, and a manifest built by the pre-upgrade toolr silently stays in use.

## Impact

Silent — no error pointing at a stale manifest, no warning that a toolr upgrade changed behaviour. The command just keeps failing (or behaving per old semantics) exactly as it did before upgrading, until someone thinks to force a manifest rebuild.

## Expected behaviour

Include the installed toolr package version (or a hash of its introspection-relevant internals) in the manifest's freshness fingerprint, so upgrading toolr alone — with zero changes to `tools/*.py` — invalidates the cached manifest and triggers a rebuild on next dispatch.

## Environment

- toolr / toolr-py, upgraded to latest (the version that includes #415)
- macOS

s0undt3ch:
## Summary

`tools/.toolr-manifest.json`'s freshness check only compared `tools/*.py`
content hashes and third-party plugin manifest hashes against the cache. A
`toolr` upgrade that changes CLI-parsing or argument-classification
behaviour (e.g. #413/#415, which reclassified some flags from
`kind: "optional"` to `kind: "repeated"`) touches neither hash, so the
stale manifest silently stays in use — with the old, wrong behaviour —
until someone thinks to run `toolr project manifest rebuild --force`.

## Fix

- Stamp `Manifest.toolr_version` with the `toolr` binary's own
  `CARGO_PKG_VERSION` on every build.
- `freshness::compare` checks it first, ahead of any hashing, so a
  mismatch never pays for a walk/glob it's going to discard.
- A mismatch always forces at least a static rebuild. It only escalates to
  a full rebuild of both layers when a tools venv is available to re-glob
  third-party fragments — escalating without one would make the
  dispatch-side rebuild discard cached third-party entries with nothing to
  repopulate them from (verified with a dedicated regression test).
- Manifests predating this field deserialize `toolr_version` as `""`,
  which never matches a real version, so they get one rebuild too.

## Testing

- `cargo test --workspace` (508 tests in `toolr-core`'s lib target alone,
  0 failures across every crate).
- `mise run test` umbrella (skill-refs drift gate + cargo test + pytest),
  clean.
- `mkdocs build --strict`, clean.
- `prek run` on all touched files, clean.
- New unit tests in `crates/toolr-core/src/freshness/tests.rs` covering:
  version mismatch + venv → full rebuild; version mismatch without a venv
  → static-only rebuild; the `""`-from-a-pre-field-manifest migration
  case.
- New integration test in `crates/toolr/tests/freshness_dispatch.rs`
  (`version_mismatch_without_venv_preserves_third_party_entries`),
  verified to actually fail if the fix is reverted.
- This diff went through three rounds of adversarial review (two by fresh
  subagents with no prior context), each of which found and fixed a real
  issue before landing clean.

Closes #420.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -42,3 +42,9 @@ no scaffolding. Just write whatever should appear in the notes.
   `toolr-py in this tools venv speaks schema N, but the toolr binary
   emitted schema M`, run `toolr project venv upgrade toolr-py` (or pin
   the `toolr` binary to a compatible version).
+- Fixed the manifest freshness check silently keeping a stale
+  `.toolr-manifest.json` after a `toolr` upgrade that changed
+  CLI-parsing/classification behaviour but touched no `tools/*.py`
+  file. The manifest now records the `toolr` version that built it, and
+  any mismatch against the running binary forces a rebuild rather than
+  trusting the cached hashes.
--- a/crates/toolr-core/src/complete/freshness.rs
+++ b/crates/toolr-core/src/complete/freshness.rs
@@ -13,7 +13,8 @@ use crate::parser::build_static_manifest;
 pub struct ResolvedManifest {
     pub manifest: Manifest,
     /// `true` if the cached on-disk manifest was returned verbatim
-    /// (both `static_hash` and `third_party_hash` matched the live state).
+    /// (`static_hash`, `third_party_hash`, and `toolr_version` all
+    /// matched the live state).
     pub from_cache: bool,
     /// The directory that contained `tools/` (the project root).
     pub project_root: PathBuf,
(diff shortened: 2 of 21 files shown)


s0undt3ch:

- #422
    - #424
        - #425 ◀

PR #428, opened 2026-08-06
fix(runner): return exit code 130 on KeyboardInterrupt instead of a raw traceback

s0undt3ch:
## What

Ctrl-C during a running command dumped a raw Python traceback and exited
with an arbitrary code instead of the conventional `130`.

## Why

`crates/toolr-core/src/execute/signals.rs` already forwards `SIGINT`/`SIGTERM`
from the parent `toolr` process to the Python runner subprocess. On the
Python side that forwarded `SIGINT` surfaces as `KeyboardInterrupt` — a
`BaseException`, not an `Exception` — so `run()`'s catch-all `except
Exception` in `_runner.py` never caught it. It fell through to CPython's
default traceback printer, and the exit code ended up being whatever the
interpreter happened to pick rather than the shell-conventional `128 + SIGINT`.

## Fix

Catch `KeyboardInterrupt` explicitly in `run()` and return `130`, ahead of
the bare `except Exception`. No traceback noise — Ctrl-C is expected user
behaviour, not a bug.

## Testing

- Added `test_run_returns_130_on_keyboard_interrupt` in
  `tests/runner/test_runner_coverage.py`, asserting exit code `130` and
  empty stderr.
- `uv run pytest tests/runner/test_runner_coverage.py -q` — 60 passed.
- Queued a release note in `UNRELEASED.md`.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -18,3 +18,8 @@ no scaffolding. Just write whatever should appear in the notes.
   days"` dependency cooldown, inherited by `tools/`, is meant for
   third-party releases; `toolr-py` is exempted from it specifically since
   it ships from the same release as the binary that runs against it.
+- Fixed Ctrl-C during a running command printing a raw Python traceback
+  and exiting with an arbitrary code instead of the conventional 130.
+  The toolr binary already forwards SIGINT to the Python runner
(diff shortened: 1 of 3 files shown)


s0undt3ch:

- #428 ◀

PR #429, opened 2026-08-06
fix(ci): exempt toolr-py from the tools/ dependency cooldown

s0undt3ch:
## What

CI's "Check if the build should run" step failed on `main`
(`toolr ci check-run-build`) with:

```
toolr runner: toolr-py in this tools venv speaks schema 2, but the toolr
binary emitted schema 3. The venv is out of sync with the binary.
```

## Why

`tools/pyproject.toml` pins `toolr-py>=0.25.0` and resolves it from PyPI.
`tools/` sits outside the root uv workspace, but it still inherits the root
`pyproject.toml`'s `[tool.uv] exclude-newer = "7 days"` dependency cooldown
when resolving. That cooldown exists to let the ecosystem vet third-party
releases before we pull them — sensible for everything else in the lock,
but `toolr-py` is our own package: it ships from the exact same release
that produces the `toolr` binary CI builds fresh off `main` HEAD to run
this check. Every schema-bumping release (like `0.27.0`, which bumped the
dispatch schema 2 → 3) therefore breaks self-CI for up to a week, until the
cooldown catches up on its own.

Confirmed by inspecting `tools/uv.lock`: locked at `toolr-py==0.26.0`
(schema 2), while `main`'s freshly-built binary is `0.27.0` (schema 3),
released the day before. The installed venv itself had drifted even
further, to `0.25.2` — two releases stale.

## Fix

Add a per-package `exclude-newer-package` override for `toolr-py` in
`tools/pyproject.toml`, exempting it from the cooldown, then re-lock and
re-sync the venv.

## Testing

- `cd tools && uv lock --upgrade-package toolr-py` — resolves to `0.27.0`.
- `cargo run -p toolr -- project venv sync` — venv updated `0.25.2 → 0.27.0`.
- `cargo run -p toolr -- ci check-run-build push main` — now exits 0
  ("Builds for the main branch should always run").
- Queued a release note in `UNRELEASED.md`.

## Note

Along the way I noticed `_runner.py`'s schema-mismatch error message tells
the user to run `toolr project venv upgrade toolr-py`, but no `upgrade`
subcommand exists (`toolr project venv sync` is correct). Left it — it's a
pre-existing, unrelated bug, filing separately rather than folding into
this PR.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,3 +10,11 @@ this file to empty for the next cycle.
 Empty between releases is the steady-state — there's no header,
 no scaffolding. Just write whatever should appear in the notes.
 -->
+
+- Fixed the `tools/` dogfooding venv lagging its own `toolr-py` pin behind
+  the `toolr` binary CI builds from `main` HEAD for up to a week after
+  every schema-bumping release, breaking self-CI (`toolr ci
+  check-run-build`) for that whole window. The root `exclude-newer = "7
+  days"` dependency cooldown, inherited by `tools/`, is meant for
+  third-party releases; `toolr-py` is exempted from it specifically since
+  it ships from the same release as the binary that runs against it.
--- a/tools/pyproject.toml
+++ b/tools/pyproject.toml
@@ -17,6 +17,17 @@ dependencies = [
     "ruamel.yaml>=0.18",
 ]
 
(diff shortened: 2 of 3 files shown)


s0undt3ch:

- #429 ◀

PR #436, opened 2026-08-14
fix(parser): classify list[T] | None / tuple[...] | None as repeatable

### LINKED ISSUE #435: list[T] | None = None parameter doesn't work — neither single nor repeated --flag values parse
**Environment:** `toolr` 0.27.0, Python 3.11.15, macOS (aarch64)

**Docs claim:** `docs/writing-commands/arguments.md` documents `list[T]` as "Annotate a parameter as `list[T]` to accept `--name VALUE` repeated multiple times (each invocation appends)", and separately documents `T | None` as the type-driven way to make a parameter optional. Combining them (`list[T] | None = None`, the natural way to express "an optional repeatable flag") does not work.

**Repro:**

```python
# tools/repro.py
from __future__ import annotations
from toolr import Context, command, command_group

command_group("repro", description="Repro commands")

@command(group="repro")
def go(ctx: Context, items: list[str] | None = None) -> None:
    """Repro for the list[T] | None bug.

    Args:
        items: items to process.
    """
    ctx.info(str(items))
```

```
$ toolr repro go
None                                          # OK - omitted flag correctly resolves to None

$ toolr repro go --items a
toolr runner: toolr runner: invalid value for `--items`: Expected `array | null`, got `str`
                                               # BUG - a single value isn't wrapped into a list

$ toolr repro go --items a --items b
error: the argument '--items <items>' cannot be used multiple times
                                               # BUG - contradicts the "each invocation appends" doc
```

**Expected:** Either (a) `list[T] | None` behaves like `list[T]` with an optional/`None` default (repeatable, defaults to `None`/`[]` when absent), or (b) the docs are corrected to say this combination isn't supported and to recommend the workaround.

**Workaround found:** `items: list[str] = ()` (any non-`None`, immutable default) works correctly for zero, one, or many `--items` occurrences.

s0undt3ch:
## Summary
- `classify_keyword_kind` peels `Annotated[...]` but never `X | None`, so
  `list[str] | None` / `tuple[...] | None` fell through to a single-value
  `Optional` kind instead of `Repeated`. One value tripped the runner's
  `array | null` type check; two values were rejected by clap as a
  repeated single-value flag.
- The type resolver already understood the union correctly
  (`Optional(List/Tuple(...))`), so the fix promotes the kind from the
  resolved type once it's known — mirroring the existing `Count` override
  in the same function.
- Also fixes a related bug in `execute_build`'s tuple early-return, which
  would otherwise send `[]` for an omitted `Optional`-wrapped tuple flag
  instead of leaving it absent for the Python `= None` default to run.

## Test plan
- [x] `cargo test --workspace` — all green
- [x] `mise run test` (skill-refs drift gate + cargo test + pytest) — all green
- [x] New regression tests: `commands.rs::optional_list_keyword_resolves_to_repeated`,
      `commands.rs::optional_tuple_keyword_resolves_to_repeated`,
      `execute_build.rs::build_spec_omits_optional_repeated_when_absent`,
      `execute_build.rs::build_spec_extracts_optional_repeated_when_present`,
      `execute_build.rs::build_spec_omits_optional_tuple_when_absent`

Fixes #435

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -23,3 +23,10 @@ no scaffolding. Just write whatever should appear in the notes.
   The toolr binary already forwards SIGINT to the Python runner
   subprocess; the runner now catches the resulting `KeyboardInterrupt`
   and exits 130 cleanly instead of leaking it as an unhandled exception.
+- Fixed `list[T] | None` and `tuple[...] | None` keyword parameters
+  being classified as single-value flags instead of repeatable ones,
+  so `--items a` failed the runner's `array | null` type check and
+  `--items a --items b` was rejected by clap as a repeated single-value
+  flag. The syntactic classifier couldn't see through the `| None`
+  union; the type resolver already could, so the fix promotes the
+  argument's kind from the resolved type once it's known.
--- a/crates/toolr-core/src/parser/commands.rs
+++ b/crates/toolr-core/src/parser/commands.rs
@@ -528,6 +528,89 @@ def check_snippets(ctx):
     }
 
     #[test]
+    fn optional_list_keyword_resolves_to_repeated() {
+        // Regression for #435: `list[T] | None` was classified as
+        // `Optional` (single-value) by the syntactic pass, since it
+        // can't see through the `| None` union — only the post-resolution
+        // override in `resolve_arguments` catches it.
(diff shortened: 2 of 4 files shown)


s0undt3ch:

- #436 ◀

PR #437, opened 2026-08-15
feat(testing): mockable ctx.run/chdir/prompt, ContextForTesting redesign

s0undt3ch:
## What

Adds first-class mocking support to `toolr.testing` for `ctx.run`, `ctx.chdir`, and `ctx.prompt`, and redesigns `make_context`'s return shape.

- **`Context` gains three injectable seams** (`_run_impl`, `_chdir_impl`, `_prompt_stream`), each defaulting to today's real behavior. `Context.run`/`chdir`/`prompt` read from these instead of hard-coded globals.
- **`toolr.testing.RunMock`** — a canonical test double for `ctx.run`. Composition over `unittest.mock.Mock` (not a `MagicMock` subclass, so a typo'd assertion raises `AttributeError` instead of silently returning a fresh mock). `.register(...)` gives `pytest-subprocess`-style declarative registration with longest-prefix matching and `occurrences` limits; `make_command_result` builds genuine `CommandResult` objects.
- **`make_context` gains `run=`/`chdir=`/`prompt_input=` params**, wiring straight into the three seams. A string `prompt_input` is wrapped in a stream that raises `EOFError` on exhaustion instead of hanging the test.
- **Breaking change:** `make_context` now returns `ContextForTesting` — a `Context` *subclass* — directly, instead of the old `ContextForTesting(ctx=, output=)` wrapper. `result.ctx.run(...)` → `ctx.run(...)`, `result.output.stdout` → `ctx.stdout`. `CapturedOutput` is removed. Migration path and rationale in `UNRELEASED.md` and the archived design doc.
- **Docs fixed**: `docs/writing-commands/testing.md` had a silently-broken example (`result.ctx.prompt = lambda ...` — raises `AttributeError` against a frozen `Context`). Rewritten, with the new mocking examples pulled in via `pymdownx.snippets` from real, CI-executed test functions instead of hand-typed code blocks, so they can't drift silently again.

## Why

Three toolr command groups in a separate caller project had each independently hand-rolled fake `Context`/`FakeRun` doubles, because there was no supported way to intercept `ctx.run`/`chdir`/`prompt` without monkeypatching internals — and the existing `make_context` docstring even said as much ("monkeypatch it yourself if a test needs to intercept it").

## How this was built

Brainstormed, spec'd, planned, and implemented via subagent-driven development (7 tasks, each independently reviewed with a fix loop where needed), followed by a final whole-branch review and three adversarial review passes. Full history in the archived design/plan:

- `specs/archive/2026/2026-08-14-testing-run-mock-design.md`
- `specs/archive/2026/2026-08-14-testing-run-mock-plan.md`

## Testing

`mise run test` (skill-refs drift gate + `cargo test --workspace` + `uv run pytest`) — 425 passed, 8 skipped (pre-existing, no built wheel in this environment), 0 failed. `prek run --all-files` clean. `mkdocs build --strict` clean, snippet includes verified to resolve against real test code.

## Follow-ups (not in this PR)

- Migrating the separate caller project's three hand-rolled fakes to `RunMock`/`make_command_result` — tracked as a follow-up in that project's own repo, once a `toolr-py` release ships this.
- A pre-existing gap surfaced during review: the repo's mypy pre-commit hook's `files:` pattern doesn't match `crates/toolr-py/python/...`, so that tree isn't actually mypy-checked by the hook. Unrelated to this change, worth its own ticket.


s0undt3ch:

- #437 ◀

PR #444, opened 2026-08-17
fix(docs): drop navigation.tabs so all top-level pages stay reachable

s0undt3ch:
Material's tab bar has no overflow menu, so with 15 top-level nav
entries the tabs past 'Third-party packages' were clipped off-screen
with no way to reach them. Nav now renders in the sidebar only, where
everything scrolls and every link is reachable.


s0undt3ch:

- #444 ◀

PR #446, opened 2026-08-18
fix(argparse): keep zero-arg Django management commands in the scanner

s0undt3ch:
The built-in argparse scanner used "found zero add_argument() calls" as
its only signal that a scanned file wasn't a real Django management
command. That also matched genuine zero-arg commands (e.g. a BaseCommand
subclass with no add_arguments at all, such as a long-running consumer),
so they were silently dropped from the dispatcher's subcommand list.

Detect Django's actual Command entry point instead: a module-level
class Command(...), or Command = <Name> aliased to a same-module class.
Files with neither the Command symbol nor any add_argument calls are
still skipped as stray helpers, and underscore-prefixed filenames
(__init__.py, _helpers.py) are now skipped up front by filename
convention rather than relying solely on the empty-args heuristic.


s0undt3ch:

- #446 ◀

PR #447, opened 2026-08-18
fix(dispatch): reject dispatcher flags leaked into a leaf's trailing args

### LINKED ISSUE #445: Dispatcher-level flags (e.g. --dry-run) silently forward to the wrapped command when placed after the subcommand name
## Summary

A dispatcher command (any command taking `dispatched: DispatchCommand`, e.g. one attached via `[tool.toolr.argparse.<name>]`) declares its own flags (`--dry-run`, `--cpu`, etc.) that must be typed *before* the dispatched subcommand name. Typed after it instead, the flag isn't rejected and isn't routed to the dispatcher — it's silently forwarded as a literal argument to the wrapped command, and the dispatcher's own parameter keeps its default value. There's no error and no warning.

This is dangerous specifically for a `--dry-run`-style safety flag: the dispatcher never sees it, so the side-effecting branch runs as if `--dry-run` had never been passed at all.

## Confirmed repro (real project, safely-observable output)

Using a `job` dispatcher wrapping Django management commands (`[tool.toolr.argparse.django]` config, attached at `jenkins.job`), with the dispatcher declaring `dry_run: bool = False`:

```
# Job option correctly placed before the subcommand — works as intended
$ toolr jenkins job --dry-run wrapup_company 12345 --stuck-company
Would trigger Jenkins job ... with parameters:
  ...
  CMD_PARAMS='manage.py wrapup_company 12345 --stuck_company'
```

```
# Same flag, placed after the subcommand — no error, no warning
$ toolr jenkins job --ram 5Gi wrapup_company 12345 --stuck-company --dry-run
Triggering Jenkins job 'pa-wrapup-company-12345-stuck-company-dry-run' on .../dashtastic ...
Started job URL: https://.../job/.../6552/
```

The triggered build's actual parameters confirm `--dry-run` was forwarded literally to the wrapped command rather than intercepted by the dispatcher:

```
CMD_PARAMS='manage.py wrapup_company 12345 --stuck-company --dry-run'
```

The dispatcher's own `dry_run` stayed `False` — it triggered the real build. This reproduces on 0.27.0 and 0.28.0.

## Why this happens (best understanding from reading the source)

- `build_dispatch_spec` (`crates/toolr/src/execute_build.rs`) splits parsing into `parent_matches` (the dispatcher's own clap-level args) and a separately packed leaf/child clap subcommand (`pack_child_args`) for the wrapped command's own scanned arguments.
- Once clap has matched the leaf subcommand name, it parses everything after it against *only* that leaf's own `Arg` set (built in `build_user_command`, `crates/toolr/src/cli.rs`). The dispatcher's flags aren't in scope there at all.
- The leaf's trailing positional (`ArgumentKind::VarPositional`, e.g. `company_ids` with `nargs='+'`) is built with `.trailing_var_arg(true)`. That's enough for clap to accept a token that doesn't match any of the leaf's own declared options as *not* an error — reconstructed faithfully via `DispatchCommand.argv` (`toolr/sources/_dispatch.py`) and forwarded to `manage.py` as-is.
- Nothing at any layer cross-checks "does this unclaimed/forwarded token happen to be a flag the *parent* dispatcher declares?" — so a job-level flag typed one position too late just silently becomes part of the wrapped command's argv.

## Why it matters beyond cosmetics

Two failure modes stack on top of each other:
1. If the wrapped command doesn't declare a `--dry-run` itself, its own argparse rejects it — the k8s job fails, but the Jenkins build (and any side effects that happen before the wrapped command's own arg-parsing runs) has already been triggered for real.
2. If the wrapped command *does* happen to declare its own `--dry-run` with unrelated semantics, it silently runs that instead — no failure, no signal that anything went wrong, and the dispatcher's own intended safety check never ran.

## Suggested fix directions (not prescriptive — happy to go whichever way fits toolr's architecture)

1. **Cross-check at spec-build time.** In `build_dispatch_spec`/`pack_child_args`, after packing the leaf's args, check whether an unclaimed/forwarded token that didn't match any of the leaf's own declared argparse arguments exactly matches a parent dispatcher flag name; error out with something like `error: '--dry-run' is an option of 'toolr jenkins job', not of 'wrapup_company' — put it before the subcommand name`.
2. **Reserve dispatcher flag names.** Refuse to scan/register a leaf argument whose flag collides with a dispatcher-level flag name, and separately, when an *unscanned* long option appears in the leftover/trailing tokens for a leaf with `trailing_var_arg`, check it against the dispatcher's flag set before accepting it as a bare value.
3. **Require `--`.** Support (or require) `toolr <dispatcher> [dispatcher-opts] -- <leaf> [leaf-opts]` so the boundary is explicit and any dispatcher flag typed after `--` is unambiguously the leaf's problem, not silently swallowed. This is the most robust but the biggest UX/back-compat change.

Happy to help narrow this down further or test a fix branch — this file lives in a private downstream project, so I've kept the example generic to the dispatcher/leaf mechanics rather than the specific management commands involved.

s0undt3ch:
A dispatcher's own flag (e.g. `--dry-run`) typed after the dispatched
subcommand name was silently swallowed by the leaf's trailing
VarPositional bucket and forwarded to the wrapped command instead of
being rejected. This left a `--dry-run`-style safety flag with no
effect at all, with no error or warning.

Closes #445

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -23,3 +23,11 @@ no scaffolding. Just write whatever should appear in the notes.
   underscore-prefixed filenames (`_helpers.py`, `__init__.py`) are now
   skipped outright for every block, Django or not — that convention is
   plain Python, not Django-specific.
+
+A dispatcher's own flag (e.g. `--dry-run` on a `job` dispatcher wrapping
+Django management commands) typed *after* the dispatched subcommand name
+was silently forwarded to the wrapped command instead of being rejected —
+the dispatcher's own parameter kept its default, so a safety flag like
+`--dry-run` had no effect at all. `toolr <dispatcher> <leaf> ... --flag`
+now errors, pointing at the correct position, instead of silently
+executing for real ([#445](https://github.com/s0undt3ch/ToolR/issues/445)).
--- a/crates/toolr/src/dispatch.rs
+++ b/crates/toolr/src/dispatch.rs
@@ -235,7 +235,7 @@ pub fn dispatch(
                     cmd.function,
                 )
             })?;
-        build_dispatch_spec(dispatcher, parent_matches, packed, &repo_root, &output_opts)
+        build_dispatch_spec(dispatcher, parent_matches, packed, &repo_root, &output_opts)?
     } else {
         build_spec(cmd, cmd_matches, &repo_root, &output_opts)
     };
--- a/crates/toolr/src/execute_build.rs
+++ b/crates/toolr/src/execute_build.rs
@@ -119,7 +119,8 @@ pub fn build_dispatch_spec(
     packed: PackedChild,
     repo_root: &Path,
     output_opts: &OutputOptions,
-) -> ExecutionSpec {
+) -> anyhow::Result<ExecutionSpec> {
+    reject_leaked_dispatcher_flags(parent, &packed)?;
     // Covers the dispatcher's own (parent) args. A dispatched child's args are
     // packed separately and not checked here — same scope as before.
     warn_relative_paths(parent, parent_matches, repo_root);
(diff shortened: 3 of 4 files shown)


s0undt3ch:

- #447 ◀

PR #450, opened 2026-08-19
fix(parser): key enum CLI choices by defining module, not bare class name

### LINKED ISSUE #449: Manifest builder resolves enum CLI choices by bare class name, causing collisions across command modules
`toolr project manifest rebuild` (and the auto-rebuild on invocation) resolves
an `Enum`'s CLI `possible values` by its bare class name (e.g. `Database`),
not its fully-qualified module path. If two unrelated command modules each
define a class named `Database` (different types, different members), the
generated manifest gives every command taking either enum the *same* choices
- whichever module was scanned last "wins" and clobbers the other's entry.

Reproduced on toolr 0.27.0 with two Python modules in the same project:

    # module_a.py
    class Database(enum.StrEnum):
        PRIMARY = "primary"
        ...

    @command("cmd-a")
    def cmd_a(ctx: Context, *, database: Database = Database.PRIMARY) -> None: ...

    # module_b.py
    class Database(enum.StrEnum):
        REPLICA = "replica"
        ...

    @command("cmd-b")
    def cmd_b(ctx: Context, *, databases: list[Database] = ()) -> None: ...

`toolr cmd-a --help` and `toolr cmd-b --help` both show the *same* list of
`possible values` - whichever of the two `Database` classes was scanned last,
even though they're unrelated types with different members. The wrong one
also silently fails at parse time if a user passes a value that's valid for
the "wrong" enum but not the type actually annotated on the command.

Worked around locally by renaming one of the two classes so there's no bare-
name collision, but the manifest builder should key on the fully-qualified
type (module + qualname), not the bare `__name__`.

s0undt3ch:
## Summary

`toolr project manifest rebuild` resolved an `Enum`'s CLI `possible
values` by its bare class `__name__`, not the module that defines it.
Two unrelated `tools/*.py` modules each declaring a same-named `Enum`
subclass (e.g. `Database`) clobbered each other's manifest entry -
whichever module the builder happened to scan last silently won for
every command referencing that name, including on the modules that
never touched.

`EnumTable` (`crates/toolr-core/src/parser/symbols.rs`) now tracks each
enum definition alongside its owning module path instead of merging
them into one flat bare-name map. Lookups resolve per the caller's own
module:

1. a same-module definition always wins,
2. failing that, an unambiguous single cross-module definition is used
   (the common "enum lives in a shared module, imported elsewhere"
   case),
3. two or more different-module definitions with no local match are
   genuinely ambiguous - the static parser doesn't track which one, if
   any, was actually imported - so this now surfaces as an
   unsupported-type build error instead of silently picking one.

Threaded the current module path through the type resolver
(`parser/types/resolve.rs`) and argument extraction
(`parser/signatures.rs`), and updated both manifest-build passes
(`parser/build.rs`, `build_fragment.rs`) to pass each file's module
path into `EnumTable::from_module`.

## Test plan

- [x] `cargo test -p toolr-core` - new unit test
      (`colliding_bare_names_resolve_per_module_not_last_merged_wins`)
      and end-to-end `build_static_manifest` regression test
      (`colliding_bare_enum_names_across_modules_resolve_independently`)
      reproducing the issue's exact two-module repro
- [x] `cargo test --workspace` - all green
- [x] `prek run --all-files` - clean (fmt, clippy, mypy, skill-refs
      drift gate)

Fixes #449

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,3 +10,10 @@ this file to empty for the next cycle.
 Empty between releases is the steady-state — there's no header,
 no scaffolding. Just write whatever should appear in the notes.
 -->
+
+- Fixed the manifest builder resolving enum CLI choices by bare class
+  name instead of by defining module, which meant two unrelated
+  `tools/*.py` modules declaring a same-named `Enum` subclass (e.g.
+  `Database`) clobbered each other's `--help` choices depending on
+  scan order. `allowed_values` and enum-attribute defaults now resolve
+  against the class actually in scope for each module ([#449](https://github.com/s0undt3ch/ToolR/issues/449)).
--- a/crates/toolr-core/src/build_fragment.rs
+++ b/crates/toolr-core/src/build_fragment.rs
@@ -97,7 +97,8 @@ pub fn build_third_party_fragment(
             path: path.clone(),
             source: e,
         })?;
-        enums.merge(EnumTable::from_module(&module));
+        let module_path = module_path_for_prefix(source_dir, path, package_name);
+        enums.merge(EnumTable::from_module(&module, &module_path));
         aliases.merge(TypeAliasTable::from_module(&module));
         sections.merge(ArgSectionTable::from_module(&module));
     }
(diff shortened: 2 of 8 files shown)


s0undt3ch:

- #450 ◀

PR #451, opened 2026-08-20
feat(testing): round out RunMock/make_context and extract the testing skill's refs from real tests

s0undt3ch:
Rounds out `toolr.testing`'s API surface and closes the process gap that let it drift out of sync with the `toolr-command-authoring` skill.

- **`RunMock`**: forwards `assert_called` and `assert_has_calls` (in addition to the already-forwarded `assert_called_once`), closing gaps in its otherwise-explicit `unittest.mock` API subset. A new completeness test diffs the forwarded set against `dir(Mock())` so a future gap fails a test instead of going unnoticed.
- **`make_context`**: accepts a `width=` override for its captured consoles, defaulting to a new `DEFAULT_TEST_CONSOLE_WIDTH` (1000 columns) so `ctx.stdout`/`ctx.stderr` assertions never have to account for rich wrapping a long line at a realistic terminal width.
- **`cargo xtask build-skill-refs`**: generalizes the existing `toolr.__all__` walker so it also covers `toolr.testing.__all__` (`references/testing.md`), and adds a new generator that extracts a real, `pytest`-run example verbatim into `references/testing-examples.md` — so the skill's worked example can never go stale or be silently wrong. `CLAUDE.md` notes the check's actual coverage boundary (per-`__all__` generators, registered by hand) so the next gap like this is a documented decision, not a silent miss.

Queued in `UNRELEASED.md`. All three `feat`s are backward-compatible additions — targets 0.30.0, not a patch release.


s0undt3ch:

- #451 ◀

PR #456, opened 2026-08-20
fix(parser): resolve command-signature enums through explicit imports

### LINKED ISSUE #454: Manifest builder rejects an enum-typed parameter imported from a different module than the one declaring it
## What happened

On 0.30.0, `toolr project manifest rebuild` hard-fails (exit 2) for three `@command`-decorated
functions in a downstream project (dashtastic), each with a plain `Enum`-typed parameter:

```
toolr: building static manifest (incl. third-party glob-merge): unsupported parameter types (3):
  - tools.metrics.analyse::analyse_data argument `env` (annotated `<expr>`): type `Environment` is not supported. Use a primitive (int, float, bool, str, pathlib.Path), a Literal[...] or Enum, or one of the aliases under `toolr.types`.
  - tools.metrics.collect::collect_data argument `env` (annotated `<expr>`): type `Environment` is not supported. Use a primitive (int, float, bool, str, pathlib.Path), a Literal[...] or Enum, or one of the aliases under `toolr.types`.
  - tools.metrics.runtime_buckets::check_runtime_buckets argument `env` (annotated `<expr>`): type `Environment` is not supported. Use a primitive (int, float, bool, str, pathlib.Path), a Literal[...] or Enum, or one of the aliases under `toolr.types`.
```

This regressed going from 0.29.0 to 0.30.0 — the same repo built its manifest cleanly (410
commands, no warnings) on 0.29.0.

## Isolated cause

The repo has three separately-declared `Environment(StrEnum)` classes across three modules
(`job/job.py`, `metrics/_common.py`, `ci/helm_diff_pr_comment.py`), all with identical members
(`PRODUCTION = "production"`, `STAGING = "staging"`) — the same same-name-collision shape
described as fixed by #449.

- `job/job.py` declares `Environment` **and** uses it as a `@command` parameter type **in the same
  file**: `env: Environment = Environment.PRODUCTION`. This resolves correctly.
- `metrics/analyse.py`, `metrics/collect.py`, `metrics/runtime_buckets.py` each **import**
  `Environment` from `metrics/_common.py` (a different module) and use it as a `@command`
  parameter: `env: Environment | None = None`. All three fail.

I narrowed it further with a reversible local edit (confirmed via `git status`/`git checkout`
before/after — not left in the tree): changing `check_runtime_buckets`'s
`env: Environment | None = None` to `env: Environment = Environment.PRODUCTION` (matching
`job/job.py`'s working shape exactly) did **not** fix it. The error just changed from
`annotated <expr>` to `annotated Environment` — same "type `Environment` is not supported"
failure. So the `X | None` shape isn't the trigger.

**The one variable that actually correlates with success/failure across all four call sites is
whether the enum class is declared in the same module as the `@command` function using it, or
imported from elsewhere.** Same-module: works. Imported from another module (with the same class
name also declared in yet other scanned modules): fails outright, even though the class itself is
a completely ordinary, resolvable `StrEnum`.

## Expectation

Given the #449 changelog note ("`allowed_values` and enum-attribute defaults now resolve against
the class actually in scope for each module"), I'd expect an *imported* enum to still resolve
correctly by following the import to the defining module — same as a type checker or `typing.get_type_hints()`
would. Right now it looks like the per-module scoping fix may only be checking the *use site's*
module for a class of that name, rather than resolving the import chain, so a same-named class
existing in another scanned module still interferes with a legitimately imported reference.

Happy to provide the actual dashtastic source (three files, none unusual) if useful, or a minimal
repro project — let me know which is more useful.

s0undt3ch:
## Summary

Fixes #454: `toolr project manifest rebuild` hard-failed on an `Enum`-typed command
parameter imported from a different module than the one declaring it, whenever any
*unrelated* module also declared a same-named enum class. The `#449` fix keyed
`EnumTable` by defining module but its cross-module fallback was a guess ("if there's
exactly one other definition, use it"), which breaks the moment a second, unrelated
module declares the same class name — even though the importing module's own `from X
import Y` statement makes the answer unambiguous.

This replaces that guess with a real import-following resolver:

- `ImportTable` tracks each module's own `from X import Y [as Z]` statements —
  absolute, relative (`from . import X`, `from .. import X`), module-top-level
  `TYPE_CHECKING` blocks, and both branches of `try`/`except ImportError`.
- `EnumTable::resolve_def` resolves same-module declarations first, then follows an
  explicit import (recursing through `__init__.py` re-exports and `as`-aliasing chains
  with a cycle/depth guard), and only falls back to the old same-name heuristic when no
  import exists at all.
- Two shapes that can never be resolved unambiguously — `from foo import *` and
  `import foo.bar` + `foo.bar.X` attribute-chain usage — now get a specific, actionable
  error instead of the generic "type is not supported" message.
- `if TYPE_CHECKING:`-guarded enum imports now work **at runtime**, not just at build
  time: confirmed empirically that `get_type_hints()` raises `NameError` on these (true
  even under `from __future__ import annotations`), so the manifest now carries the
  enum's resolved declaring module, and the Python runner lazily imports the real class
  into `get_type_hints`'s `localns` at coercion time — after all modules have finished
  their own import phase, so it doesn't reintroduce the circular-import problem
  `TYPE_CHECKING` exists to dodge.
- Also fixed along the way: `Environment | None`-style `Optional`-wrapped `Enum`/
  `Literal` parameters built successfully but silently populated `allowed_values` as
  empty — no clap choice validation, nothing in `--help`.

Design + plan (archived): `specs/archive/2026/2026-08-20-command-symbol-resolution-{design,plan}.md`

## Test plan

- [x] `cargo test --workspace` — 549+ Rust tests, including new coverage for relative
      imports, `__init__.py` re-export chains, `as`-aliasing, `try`/`except` dual-path
      imports, cycle guards, and the star-import/attribute-chain rejection messages.
- [x] `uv run pytest` — 438 passed, including new `_coerce_args` coverage for the
      `TYPE_CHECKING`-only-import shape with and without `enum_modules` supplied.
- [x] End-to-end integration test (`crates/toolr/tests/freshness_dispatch.rs`) drives
      the real `toolr` binary against the exact reported shape and asserts the
      persisted manifest carries both the resolved `allowed_values` and the enum's
      actual declaring module.
- [x] `cargo xtask build-skill-refs --check` — no drift.
- [x] Full `mise run test` umbrella — green.


s0undt3ch:

- #456 ◀

PR #457, opened 2026-08-20
docs(claude): document version-bump judgment and regression-test discipline

s0undt3ch:
Two process notes learned during the #454 fix (PR #456):

1. `tools/version.py::bump` takes an explicit version — it doesn't compute a semver bump from commit history. Documents that the minor-vs-patch call should come from reading `UNRELEASED.md`'s content together with the commit log, not from Conventional Commit prefixes alone.
2. A regression-test fixture built from a bug report that fails on something unrelated to the targeted bug is a second finding, not friction to simplify away — codifies a near-miss from this session where weakening a fixture almost hid an independent bug.


s0undt3ch:

- #457 ◀

PR #459, opened 2026-08-21
fix(renovate): split toolr-py versioning from mise/gh-actions regex

s0undt3ch:
The shared 'toolr release' packageRule applied a strict major.minor.patch
regex versioning to pep621, mise, and github-actions together. That regex
can't parse the '>=0.25.0' range in tools/pyproject.toml, so Renovate
silently stopped bumping toolr-py in tools/uv.lock after the versioning
override landed — mise/action bumps kept working, the pep621 lockfile
bump didn't. Split into two rules under the same groupName: pep621 keeps
default PEP 440 versioning, mise/github-actions keep the regex override.


s0undt3ch:

- #459 ◀

PR #460, opened 2026-08-21
ci(matrix): narrow PR Windows/macOS tests to latest Python only

s0undt3ch:
Linux keeps the full TEST_PYTHONS range on every build. Windows and macOS runner-minutes cost more, so PR builds now test only the latest CPython there; the full range still runs whenever `generate_build_matrix` resolves to release mode (push to `main`, or a `full-build`-labelled PR).


s0undt3ch:

- #460 ◀

